PR #12, opened 2023-08-08
Refactor jupyterhub and add Makefile, gitignore

chiayi:
**What is this PR for:**
Refactors the `jupyterhub` module part out of the `modules/` directory. This way, user can separately install jupyterhub if they already have a cluster running. Added instructions in the README. 
Also added a Makefile that can be used but I would still consider it WIP. 
Added a gitignore file to ignore the terraform state files.


ray-on-gke/user/jupyterhub/jupyter_config/config.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+scheduling:
+  userScheduler:
+    enabled: false
+
+singleuser:
+  nodeSelector:
+    iam.gke.io/gke-metadata-server-enabled: "true"
+  image:
+    name: jupyter/datascience-notebook
```
imreddy13, line 47: Why this image?

ray-on-gke/user/jupyterhub/jupyter_config/config.yaml
```diff
@@ -0,0 +1,78 @@
...
+    tag: python-3.10
+  startTimeout: 700
+  profileList:
+    - display_name: "Prof1"
```
imreddy13, line 51: What are we trying to achieve with these profiles?
Maybe we could have 1 CPU, 1 GPU, 1 TPU example?
And also illustrate how to set memory limits?

ray-on-gke/user/jupyterhub/jupyter_config/config.yaml
```diff
@@ -0,0 +1,78 @@
...
+      description: "Profile 1"
+      default: true
+    - display_name: "Prof2"
+      description: "Profile 2 with extra gpus"
+      kubespawner_override:
+        image: jupyter/datascience-notebook:python-3.11
```
richardsliu, line 57: Why is this python 3.11 when the default one is 3.10?

ray-on-gke/user/jupyterhub/variables.tf
```diff
@@ -15,5 +15,11 @@
 variable "namespace" {
   type        = string
   description = "Kubernetes namespace where resources are deployed"
-  default     = "ray"
+  default     = "<your user name>"
+}
+
+variable "enable_create_namespace" {
```
richardsliu, line 21: Just "create_namespace".

Makefile
```diff
@@ -0,0 +1,84 @@
...
+	cd ./ray-on-gke/user/ && terraform destroy -auto-approve -var namespace=$(NAMESPACE) -var service_account=$(SA-ACCOUNT-NAME)
+
+delete-jupyterhub:
+	cd ./ray-on-gke/user/jupyterhub && terraform destroy -auto-approve
+
+clean-tfstate: TF-DIR ?= platform
+clean-tfstate:
+	cd ./ray-on-gke/$(TF-DIR)/ && rm terraform.tfstate terraform.tfstate.backup
+
+test-stuff:
```
richardsliu, line 83: Is this needed?

ray-on-gke/user/jupyterhub/jupyter_config/config.yaml
chiayi: There isn't a specific reason I picked this image. I will most likely change this to `<image-name>` or somehow use helm charts to fill in the variable?

Makefile
chiayi: I don't think this will ever be needed if the config path is set properly, but I added it just in case the fix requires a terraform state reset.

chiayi:
I will go through the jupyter yaml and set some of the values to the `<user-input-here>` style.

ray-on-gke/user/jupyterhub/jupyter_config/config.yaml
chiayi: How do you request for a TPU? currently there is a request/limit field set in the profiles to get the gpus.

ray-on-gke/user/jupyterhub/jupyter_config/config.yaml
chiayi: Changed to just <image-name>:<tag>

ray-on-gke/user/jupyterhub/variables.tf
chiayi: done1

ray-on-gke/user/jupyterhub/jupyter_config/config.yaml
chiayi: Changed it to \<image-name\> and \<tag\>

ray-on-gke/user/jupyterhub/jupyter_config/config.yaml
richardsliu: TPUs are not supported yet (pending @yiyinglovecoding 's PR)

richardsliu (approved):

imreddy13 (approved):

PR #22, opened 2023-08-30
Setup ray head fault tolerance with Redis

imreddy13:
Add fault tolerance to the ray cluster via Redis so the GCS state is persisted.

The ray cluster has no fault tolerance by default i.e. if the head node dies, the application is disrupted because all the worker nodes restart. Additionally, job logs stored in the ray dashboard are lost.

The following changes enable fault tolerance:
1) Deploy a Redis service in the GKE cluster.
2) Modify the kuberay config with annotations that enable fault tolerance and set environment variable `RAY_REDIS_ADDRESS` to the Redis service.

This can be tested by creating a ray cluster with fault tolerance enabled, then force deleting the head node pod and verifying the worker pod does not restart.


spencer-p:
Closing as stale/merge conflicts

PR #32, opened 2023-09-08
add tpu node location in template

yiyinglovecoding:
TPU V4 is only available in us-central2-b. Otherwise everytime we need to manually add this in template so can pass through. 

And a bit auto formatting.

### DIFF
--- a/ray-on-gke/platform/modules/gke_standard/variables.tf
+++ b/ray-on-gke/platform/modules/gke_standard/variables.tf
@@ -52,3 +52,9 @@ variable "enable_tpu" {
   description = "Set to true to create TPU node pool"
   default     = false
 }
+
+variable "tpu_node_locations" {
+  type        = list(string)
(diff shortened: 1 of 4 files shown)


ray-on-gke/platform/modules/gke_standard/variables.tf
```diff
@@ -51,4 +51,10 @@ variable "enable_tpu" {
   type        = bool
   description = "Set to true to create TPU node pool"
   default     = false
+}
+
+variable "tpu_node_locations" {
```
richardsliu, line 56: Do we need this as a variable? Since availability is limited, maybe it would be sufficient to just hardcode it?

yiyinglovecoding: yes we can also do hardcode

richardsliu (approved):

PR #39, opened 2023-09-14
serving Llama 70b on 2 x  L4 GPU

samos123:
Tutorial on how to use text-generation-inference with Llama 2 70B Chat


serving-llama2-70b-on-l4-gpus/text-generation-interface.yaml
```diff
@@ -0,0 +1,48 @@
...
+          limits:
+            nvidia.com/gpu: 2
+        env:
+        - name: MODEL_ID
+          value: meta-llama/Llama-2-70b-chat-hf
+        - name: NUM_SHARD
+          value: "2"
+        - name: PORT 
+          value: "8080"
+        - name: QUANTIZE
```
danielvegamyhre, line 28: It might be worth adding a short sentence explaining what quantization is and why we are doing it for inference, with a link to more detailed information if the user is interested in learning more.

serving-llama2-70b-on-l4-gpus/text-generation-interface.yaml
```diff
@@ -0,0 +1,48 @@
...
+      containers:
+      - name: llama-2-70b
+        image: ghcr.io/huggingface/text-generation-inference:1.0.3
+        resources:
+          limits:
+            nvidia.com/gpu: 2
+        env:
+        - name: MODEL_ID
+          value: meta-llama/Llama-2-70b-chat-hf
+        - name: NUM_SHARD
```
danielvegamyhre, line 24: It might be helpful to give a brief bullet point on what some of these less self-explanatory environment variables do, such as `NUM_SHARD`, so people can adapt this to other inference infra setups. For example, if you try googling `text-generation-interface \"NUM_SHARD\"` I don't see any relevant docs about this env var (maybe I'm just not finding them), and so the user will be forced to read the [comments](https://github.com/huggingface/text-generation-inference/blob/v1.0.3/launcher/src/main.rs#L119-L124) in the main.rs file for answers, which is not ideal.

danielvegamyhre:
@samos123 Added a few minor comments but otherwise LGTM!

serving-llama2-70b-on-l4-gpus/text-generation-interface.yaml
samos123: I didn't find any good blogs that I would link to so I rather leave that up to the user to figure out. I did add some comments on why we use quantization and explained it.

serving-llama2-70b-on-l4-gpus/text-generation-interface.yaml
samos123: I updated with some basic explanation.

samos123:
@danielvegamyhre could you please review again?

danielvegamyhre:
/lgtm

cc @richardsliu who has approval powers in this repo

serving-llama2-70b-on-l4-gpus/text-generation-interface.yaml
ahg-g: I think we should at least explain why we selected this value.

samos123: I added an explanation already and also added explanation on how this setting impacts how much GPU memory is needed

richardsliu (approved):

tutorials/serving-llama2-70b-on-l4-gpus/README.md
```diff
@@ -0,0 +1,152 @@
...
+           emptyDir:
+              medium: Memory
+         - name: data
+           hostPath:
+            path: /mnt/stateful_partition/kube-ephemeral-ssd/llama-data
+      nodeSelector:
+        cloud.google.com/gke-accelerator: nvidia-l4
+```
+
+Inside the YAML file the following settings are used:
```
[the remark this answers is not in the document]
samos123: weird, I'm not sure why I never found that doc before. I've added the link to the PR.

PR #48, opened 2023-09-18
Adding Authentication for Jupyterhub 

chiayi:
Still WIP, as I need to create a GCR repo that will have the correct images and a few other things.


chiayi:
Going to split the PR into multiple smaller PRs

PR #60, opened 2023-10-04
[cloudbuild] add ray api call

yiyinglovecoding:
- add ray job submit command in the script 
- add new folder `example_ray_job_scripts` under `ray-on-gke` for test ray python file


BogdanAtWork (approved):
LGTM

PR #73, opened 2023-10-13
Changed order of resource creation/installation

chiayi:
The order of Jupyter installation was causing issues with Autopilot.


jupyter-on-gke/iap_module/output.tf
imreddy13, line 1: not sure why this appears in the diff?

jupyter-on-gke/jupyterhub.tf
```diff
@@ -31,6 +31,15 @@ provider "google-beta" {
   region = var.location
 }
 
+data "google_compute_backend_service" "jupyter-ingress" {
```
imreddy13, line 34: Could you add comments to each resource we create explaining why we need it/how it's used by IAP?

imreddy13 (approved):

jupyter-on-gke/iap_module/output.tf
chiayi: Oh because I deleted the file. It was exporting the service unique id, but I moved it out since it would create a circular dependency.

jupyter-on-gke/jupyterhub.tf
chiayi: Added some comments

PR #82, opened 2023-10-17
Allow user provided domain

chiayi:


jupyter-on-gke/iap_module/variables.tf
```diff
@@ -49,4 +49,16 @@ variable "client" {
     secret = string
   })
   description = "oauth client"
+}
+
+variable "url_domain_addr" {
```
imreddy13, line 54: Probably also needs a readme update.
I think we need a separate PR that either documents all the vars/options in the readme or in a different .md file referenced in the readme.

jupyter-on-gke/iap_module/variables.tf
```diff
@@ -49,4 +49,16 @@ variable "client" {
...
   description = "oauth client"
+}
+
+variable "url_domain_addr" { 
+  type = string
+  description = "Domain provided by the user. If it's empty, we will create one for you."
+  default = ""
+}
+
+variable "url_domain_name" {
```
imreddy13, line 60: What is name vs address? Where is name used?

jupyter-on-gke/iap_module/iap.tf
```diff
@@ -44,19 +45,19 @@ resource "kubectl_manifest" "backend_config" {
...
+  depends_on = [ kubernetes_secret.my-secret ]
 }
 
 # Ingress for IAP
 resource "kubectl_manifest" "static_ingress" {
   override_namespace = var.namespace
 
   yaml_body          = templatefile("${path.module}/deployments/static-ingress.yaml", {
-    static_addr_name = "${google_compute_global_address.default.name}"
+    static_addr_name = var.url_domain_addr != "" ? var.url_domain_name : "${google_compute_global_address.default[0].name}"
```
imreddy13, line 58: should this be url_domain_name?

jupyter-on-gke/iap_module/variables.tf
chiayi: Yeah definitely, I have a task for myself to add a definitions/"meta" file.

jupyter-on-gke/iap_module/variables.tf
chiayi: So the name is used by the ingress annotations. It's used to specify the global ip address.

jupyter-on-gke/iap_module/iap.tf
chiayi: So we don't want it to use the domain name if we are reserving an IP for them. I specified in the variable description that the domain name will only be used if they provide a domain url.

jupyter-on-gke/iap_module/variables.tf
chiayi: I added a note to the readme as well.

imreddy13 (approved):

PR #83, opened 2023-10-17
Automatically add principals as web users from allowlist file

chiayi:
Add all the users/groups specified in the `allowlist` file as web app users for IAP.

### DIFF
--- a/jupyter-on-gke/jupyterhub.tf
+++ b/jupyter-on-gke/jupyterhub.tf
@@ -41,6 +41,14 @@ data "google_compute_backend_service" "jupyter-ingress" {
   project = var.project_id
 }
 
+resource "google_iap_web_backend_service_iam_binding" "binding" {
+  count = var.add_auth && data.google_compute_backend_service.jupyter-ingress.generated_id != null ? 1 : 0
+  project = var.project_id
+  web_backend_service = var.add_auth && data.google_compute_backend_service.jupyter-ingress.generated_id != null ? "${data.google_compute_backend_service.jupyter-ingress.name}" : "no-id-yet"
+  role = "roles/iap.httpsResourceAccessor"
+  members = split("\n", chomp(file("${path.module}/allowlist")))
+}
+
 # Enabled the IAP service 
 resource "google_project_service" "project_service" {
   count = var.enable_iap_service ? 1 : 0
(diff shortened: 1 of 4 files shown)


jupyter-on-gke/jupyterhub.tf
```diff
@@ -41,6 +41,14 @@ data "google_compute_backend_service" "jupyter-ingress" {
   project = var.project_id
 }
 
+resource "google_iap_web_backend_service_iam_binding" "binding" {
+  count = var.add_auth && data.google_compute_backend_service.jupyter-ingress.generated_id != null ? 1 : 0
```
imreddy13, line 45: data.google_compute_backend_service.jupyter-ingress.generated_id != null --> why do we need this?

jupyter-on-gke/variables.tf
```diff
@@ -70,4 +70,10 @@ variable "url_domain_name" {
   type = string
   description = "Name of the domain provided by the user. This var will only be used if url_domain_addr is not empty"
   default = ""
+}
+
+variable "support_email" {
+  type = string
+  description = "Email displayed on the OAuth consent screen."
```
imreddy13, line 77: Maybe we can add a Pantheon link to where this email can be found?

jupyter-on-gke/jupyterhub.tf
imreddy13: What happens if the backend service doesn't exist yet? Does the terraform apply fail?

chiayi: Yup, it gives an 404 error in terraform saying that the service doesn't exist.

jupyter-on-gke/variables.tf
chiayi: Oh wait, this is part of the automatically creating clients, required field for creating a brand. Let me know if you would like me to create another PR for this. 

But without a brand, they won't see the email on the Pantheon page.

chiayi: I'll add a better description, not exactly telling you what it is.

jupyter-on-gke/variables.tf
```diff
@@ -70,4 +70,10 @@ variable "url_domain_name" {
   type = string
   description = "Name of the domain provided by the user. This var will only be used if url_domain_addr is not empty"
   default = ""
+}
+
+variable "support_email" {
+  type = string
+  description = "Email for users to contact with questions about their consent"
```
imreddy13, line 77: Still think that the admin setting up this environment needs to know where to find this email. Can we provide either a link to the Pantheon oauth consent page where they can find it, or instructions for how to find it?

imreddy13 (approved):

jupyter-on-gke/variables.tf
imreddy13: Ohh, I see - we just associate whatever email they provide with the brand we create? In which case, ignore this :)

PR #88, opened 2023-10-20
Combine jupyterhub yaml and also some screenshots for readme

chiayi:
Only one config is needed for Jupyterhub yaml. Depending on if auth is enabled it will use the dummy auth or GCP IAP


jupyter-on-gke/jupyter_config/singleprofile-config.yaml
imreddy13, line 1: Don't we need to delete config.yaml too?

jupyter-on-gke/jupyter_config/config-selfauth.yaml
```diff
@@ -12,15 +12,25 @@ hub:
...
     tag: latest
-  config: 
+  config:
     JupyterHub:
-      authenticator_class: 'gcpiapjwtauthenticator.GCPIAPAuthenticator'
+      authenticator_class: ${authenticator_class}
     GCPIAPAuthenticator: 
       header_name: 'X-Goog-IAP-JWT-Assertion'
       expected_audience: '/projects/${project_number}/global/backendServices/${service_id}'
+    Authenticator:
```
imreddy13, line 21: Did we test that these credentials do not work if auth/IAP is enabled, with the current config?

chiayi: Yes, these username/password are not valid Google username/passwords.

imreddy13 (approved):

PR #89, opened 2023-10-23
[cloud build] Test jupyterhub function

yiyinglovecoding:
add function test for terraform template under `jupyter-on-gke/`
https://screenshot.googleplex.com/3CkvN6E7SnZr3WD
- add `jupyter-on-gke/tests/change_jupyter_config.py` for change the image name in `config-selfauth.yaml`
   - current user experience is building and pushing the image by themselves. For cloudbuild, the image is ready in project `gke-ai-eco-dev`, and `change_jupyter_config.py` will replace `<Repo name here>` with the actualy image name
- add `jupyter-on-gke/tests/test_hub.py` for testing if jupyterhub is up and running 
- add `test jupyterhub` step in `cloudbuild.yaml`
   -  currently set `add_auth` to false, and testing hub will look at `proxy-public` url

extra change: nit: tf format


ray-on-gke/cloudbuild.yaml
```diff
@@ -64,30 +67,82 @@ steps:
...
+          echo "ray API run failed"
           exit 1
         fi
 
         if [[ $(cat /workspace/user_result.txt) == "false" ]]; then
+          echo "ray cluster failed"
+          exit 1
+        fi
+
+        if [[ $(cat /workspace/jupyterhub_tf_result.txt) == "false" ]]; then
```
andrewsykim, line 136: Instead of failing if `/workspace/jupyterhub_tf_result.txt` has `false`, could we instead check if this has `pass` and fail in all other cases? That seems less prone to false negatives (such as the typo on line 59  where `/workspace/user_result` should have been `/workspace/user_result.txt`)

ray-on-gke/cloudbuild.yaml
```diff
@@ -64,30 +67,82 @@ steps:
...
+        -var=enable_iap_service=false \
+        -auto-approve -no-color 
+
+        cd ../ray-on-gke/user/
+        terraform destroy -var=project_id=$PROJECT_ID \
+        -var=namespace=$SHORT_SHA \
+        -var=service_account=$_USER_NAME-$SHORT_SHA-system-account \
+        -auto-approve -no-color
+
+        gcloud container clusters delete ray-$SHORT_SHA-$_PR_NUMBER-cluster --zone $_ZONE  --project $PROJECT_ID --quiet
```
andrewsykim, line 119: why not terraform destry in gke-platform instead of running `gcloud container clusters delete`?

ray-on-gke/cloudbuild.yaml
```diff
@@ -64,30 +67,82 @@ steps:
...
+        -auto-approve -no-color \
+        || ( echo "false" > /workspace/jupyterhub_tf_result.txt )
+
+        kubectl wait --all pods -n $SHORT_SHA --for=condition=Ready --timeout=300s
+        kubectl get services -n $SHORT_SHA
+        kubectl get service proxy-public -n $SHORT_SHA --output jsonpath='{.status.loadBalancer.ingress[0].ip}' > /workspace/jupyterhub_host_url.txt
+        echo "HOST URL is " $(cat /workspace/jupyterhub_host_url.txt)
+
+        cd tests/
+        python3 test_hub.py $(cat /workspace/jupyterhub_host_url.txt) || ( echo "false" > /workspace/jupyterhub_test_result.txt )
```
andrewsykim, line 96: Question since I'm not super familiar with cloud build, but is tracking the test result in `/workspace/jupyterhub_test_result.txt ` needed? Shouldn't `python3 test_hub.py $(cat /workspace/jupyterhub_host_url.txt)` returning exit code 1 be enough for this stage to be marked failed?

jupyter-on-gke/tests/test_hub.py
```diff
@@ -0,0 +1,64 @@
...
+    Tests the hub api's root endpoint (/). The hub's version should be returned.
+
+    A typical jupyterhub logging response to this test:
+
+        [I 2019-09-25 12:03:12.051 JupyterHub log:174] 200 GET /hub/api (test@127.0.0.1) 9.57ms
+    """
+    r = requests.get(hub_url + "/hub/api")
+    r.raise_for_status()
+    info = r.json()
+    assert V("4") <= V(info["version"]) <= V("5")
```
andrewsykim, line 25: What's the expected value of `info["version"]` here? Is `V("4")` the major version?

yiyinglovecoding: Yes 4 is the major version, the response is 
```
{'version': '4.0.2'}
```

ray-on-gke/cloudbuild.yaml
yiyinglovecoding: Here is I confronted a error for terraform destroy saying cluster's deletion protection is enabled, I tried to find where it came from but failed. So I changed to gcloud command and it went through and I think this should delete everything(?..)
https://screenshot.googleplex.com/8Gh9525zozof4uD

ray-on-gke/cloudbuild.yaml
yiyinglovecoding: done

ray-on-gke/cloudbuild.yaml
andrewsykim: I saw you updated the checks to:
```
python3 test_hub.py $(cat /workspace/jupyterhub_host_url.txt) || ( exit 1 )
```

Will that prevent clean up of the cluster then?

andrewsykim: ah sorry, I missed the `allowExitCodes: 1` change

ray-on-gke/cloudbuild.yaml
```diff
@@ -63,31 +69,88 @@ steps:
...
+
+        cd ../ray-on-gke/user/
+        terraform destroy -var=project_id=$PROJECT_ID \
+        -var=namespace=$SHORT_SHA \
+        -var=service_account=$_USER_NAME-$SHORT_SHA-system-account \
+        -auto-approve -no-color
+
+        gcloud container clusters delete ray-$SHORT_SHA-$_PR_NUMBER-cluster --zone $_ZONE  --project $PROJECT_ID --quiet 
+
+        if [[ $(cat /workspace/gke_cluster_result.txt) != "pass" ]]; then
```
andrewsykim, line 129: Is it possible to put these checks into a separate build stage instead of the clean up one? This way you can set `allowFailure: true` in the other stages (instead of allowExitCodes) except the final stage that checks these values

andrewsykim: (and it's a little misleading to see the failure in the clean up stage since it won't be immediately clear the build failed due to the test and not due to the clean up)

ray-on-gke/cloudbuild.yaml
yiyinglovecoding: I see where the deletion_protection come up. This is relatively new on terraform google provider that it will automatically set `deletion_protection=true` for gke cluster.
According to doc,
`On version 5.0.0+ of the provider, you must explicitly set deletion_protection=false (and run terraform apply to write the field to state) in order to destroy a cluster. It is recommended to not set this field (or set it to true) until you're ready to destroy.`
So to do it in my cloudbuild I need to make `deletion_protection` a variable, and set it to false first and then destroy. but I am hesitate of introducing a new variable to the current template that is not for customers.

jupyter-on-gke/tests/test_hub.py
```diff
@@ -0,0 +1,64 @@
...
+    print("Jupyterhub Rest API is working.")
+
+
+def test_hub_login(hub_url):
```
imreddy13, line 29: @yiyinglovecoding As we discussed offline, can we create a separate tracking bug to test the actual IAP auth flow, once you separate the ray and jupyter tests? FYI the current blocker is that provisioning the managed cert for IAP takes 20-30 mins, but since this is a presubmit test, it should be ok to wait for the cert to provision.

ray-on-gke/cloudbuild.yaml
yiyinglovecoding: done

jupyter-on-gke/tests/test_hub.py
yiyinglovecoding: will do in separate PR

ray-on-gke/cloudbuild.yaml
```diff
@@ -26,34 +34,32 @@ steps:
...
+        --project $PROJECT_ID 
+
+        cd ray-on-gke/user/
         terraform apply \
         -var=project_id=$PROJECT_ID \
         -var=namespace=$SHORT_SHA \
         -var=service_account=$_USER_NAME-$SHORT_SHA-system-account \
         -auto-approve -no-color \
-        || ( echo "false" > /workspace/user_result )
+        || ( exit 1 )
```
andrewsykim, line 61: is the `|| ( exit 1)` step still needed if we have `allowFailure: true`?

andrewsykim (commented):
LGTM, deferring approval to @chiayi or @imreddy13

ray-on-gke/cloudbuild.yaml
yiyinglovecoding: Good catch

imreddy13 (approved):

PR #97, opened 2023-10-26
install kuberay-operator in a dedicated namespace (ray-system)

andrewsykim:
I got the impression from existing variables in the kuberay module that we intended to install kuberay-operator in the `ray` namespace, but since the namespace definition was missing from `helm_release` we are still installing kuberay-operator in the `default` namespace. This PR adds the namespace field to the helm release of kuberay-operator and changes the default value for the namespace to `ray-system`.

### DIFF
--- a/gke-platform/modules/kuberay/variables.tf
+++ b/gke-platform/modules/kuberay/variables.tf
@@ -27,7 +27,7 @@ variable "cluster_name" {
 variable "namespace" {
   type        = string
   description = "Kubernetes namespace where resources are deployed"
-  default     = "ray"
+  default     = "ray-system"
 }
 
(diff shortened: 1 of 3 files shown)


gke-platform/modules/kuberay/variables.tf
```diff
@@ -27,7 +27,7 @@ variable "cluster_name" {
 variable "namespace" {
   type        = string
   description = "Kubernetes namespace where resources are deployed"
-  default     = "ray"
+  default     = "ray-system"
```
andrewsykim, line 30: kuberay-system instead?

richardsliu (approved):

PR #99, opened 2023-10-31
Example for inference using a single TPU

xiangshen-dk:
Added an example (MINST) which is used for a quickstart tutorial.


alizaidis:
@ahg-g Any chance we can get a review on this example showing how to train and do batch inference using MNIST on a single TPU VM? Much appreciated!

gke-tpu-examples/training/mnist-single-tpu/src/gke-config/standard-tensorflow-bash-v4.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: v1
+kind: Pod
```
ahg-g, line 2: We shouldn't recommend creating plain pods, why not a job?

gke-tpu-examples/training/mnist-single-tpu/src/gke-config/standard-tf-mnist-batch-predict-v4.yaml
```diff
@@ -0,0 +1,40 @@
...
+      nodeSelector:
+        cloud.google.com/gke-tpu-topology: 2x2x1  # target topology
+        cloud.google.com/gke-tpu-accelerator: tpu-v4-podslice
+      containers:
+      - name: tensorflow
+        image: tensorflow/tensorflow:2.14.0
+        securityContext:
+          privileged: true
+        command: ["/bin/bash", "-c", "--"]
+        args: ["cd /data/tensorflow-mnist-example; pip install -r requirements.txt; python tensorflow_mnist_batch_predict.py"]
```
ahg-g, line 21: where do we create the container that includes the `tensorflow_mnist_batch_predict.py` script?

gke-tpu-examples/training/mnist-single-tpu/src/gke-config/standard-tensorflow-bash-v4.yaml
xiangshen-dk: That's the manifest to create a pod so the user can have a shell session to test the volume and verify the TPU attachment. The real training and inferencing are using jobs.

gke-tpu-examples/training/mnist-single-tpu/src/gke-config/standard-tf-mnist-batch-predict-v4.yaml
xiangshen-dk: It should be on the TPU node based on the node selector.

ahg-g: oh, the script is accessed from gcs; ok, this is a slightly strange setup, usually we put it inside the container

ahg-g (approved):

PR #104, opened 2023-11-01
fix: replace kubectl provider with kubernetes provider

umeshkumhar:


imreddy13 (approved):

gke-platform/main.tf
```diff
@@ -77,21 +69,21 @@ module "gke_standard" {
...
-  depends_on       = [module.gke_autopilot, module.gke_standard]
-  region           = var.region
-  cluster_name     = var.cluster_name
-  enable_autopilot = var.enable_autopilot
-}
+#   depends_on       = [module.gke_autopilot, module.gke_standard]
+#   region           = var.region
+#   cluster_name     = var.cluster_name
+#   enable_autopilot = var.enable_autopilot
+# }
```
tallclair, line 89: Why is this all commented out? Should it be deleted?

gke-platform/modules/kubernetes/kubernetes.tf
```diff
@@ -16,7 +16,7 @@ data "http" "nvidia_driver_installer_manifest" {
   url = "https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/nvidia-driver-installer/cos/daemonset-preloaded.yaml"
 }
 
-resource "kubectl_manifest" "nvidia_driver_installer" {
-  yaml_body = data.http.nvidia_driver_installer_manifest.response_body
-  count = var.enable_tpu || var.enable_autopilot ? 0 : 1
+resource "kubernetes_manifest" "nvidia_driver_installer" {
```
tallclair, line 19: How does this work? My understanding was that `kubernetes_manifest` wouldn't work if the cluster was not yet created, but the cluster is also being created as part of the gke-platform module, right?

tallclair: I see, it looks like this is now being called by the ray user template, rather than the `gke-platform/main.tf`? In that case, I think this module (and maybe also the kuberay one?) should be moved out of the gke-platform directory.

gke-platform/main.tf
```diff
@@ -77,21 +69,21 @@ module "gke_standard" {
   enable_tpu       = var.enable_tpu
 }
 
-module "kubernetes" {
-  source = "./modules/kubernetes"
+# module "kubernetes" {
```
richardsliu, line 72: Is this intended to be removed?

ray-on-gke/user/main.tf
```diff
@@ -18,16 +18,23 @@ provider "kubernetes" {
...
+module "kubernetes-nvidia" {
+  source           = "../../gke-platform/modules/kubernetes"
+  enable_autopilot = var.enable_autopilot
+  enable_tpu       = var.enable_tpu
+}
+
+module "kuberay-operator" {
+  source           = "../../gke-platform/modules/kuberay"
+  enable_autopilot = var.enable_autopilot
+}
```
tallclair, line 36: These are pieces that we expect to be installed by the cluster admin, but my understanding of the division was that the user module was intended to be installed by non-admin users?

I think what we may need to do is split gke-platform into 2 pieces: one to create the cluster & node pools, etc; and a second phase to set up the cluster-wide kubernetes workloads (kuberay operator & nvidia drivers)

tallclair (commented):
Thanks for making this change.  My understanding of terraform is limited, so let me know if I'm way off base with these comments!

gke-platform/main.tf
umeshkumhar: yes it can be deleted. I keep this for testing.. but yeah let me remove that

gke-platform/modules/kubernetes/kubernetes.tf
umeshkumhar: So my understanding is when we say platform, its mostly the base setup required in terms of GCP services, something like, provision, GKE, nodepools, IAM, IAP, etc..

and the workloads (or applications) is something which will be deployed on kubernetes engine as container/services/ingress/etc. 

So, ideal pipeline would be like:
1. provision platform:  provision GKE with nodepools, setup required IAM, enable APIs, etc.
2. Provision workloads: now GKE is ready, deploy all workloads like ray-operator, ray-cluster, jupyter-hub, stable-difussion, etc.. using kubernetes providers.

In that way the `kubernetes_manifest` will also work as GKE got prepared in first step.

ray-on-gke/user/main.tf
umeshkumhar: yes right, in that case we should have one separate section, or other way can be to convert manifests into simple helm charts and use `helm_release` instead..

if we want to go through user personas way, then we first need to define what is expected out of these users and accordingly may be create two sub folders inside workloads..  but we can brainstorm more on this on call. Feel free to schedule one.

PR #105, opened 2023-11-01
Add licensing and change contribution md location

chiayi:
This is required as part of launch. Having a launch is required to setup a public project.


tallclair:
This should have a pre-submit check if we want to maintain the requirement. You can look into reusing the Kubernetes script if there isn't an off the shelf github version: https://github.com/kubernetes/kubernetes/tree/master/hack/boilerplate

imreddy13:
> This should have a pre-submit check if we want to maintain the requirement. You can look into reusing the Kubernetes script if there isn't an off the shelf github version: https://github.com/kubernetes/kubernetes/tree/master/hack/boilerplate

imreddy13 (approved):
+1 that this needs a presubmit so we don't ever have to do this again, but I think that can be a separate task to unblock the turbo

PR #111, opened 2023-11-06
pin ray-cluster helm chart version to 0.6.1

andrewsykim:
In https://github.com/GoogleCloudPlatform/ai-on-gke/pull/81 and https://github.com/GoogleCloudPlatform/ai-on-gke/pull/96 we pinned the kuberay-operator image version and the helm chart version to v0.6.1. However, the ray-cluster chart which installs the default RayCluster is using the latest version (v1.0.0-rc.1) which is trying to use the new v1 version of the CRDs that are not supported in v0.6.1 yet.


ryanaoleary (approved):
LGTM

imreddy13 (approved):

PR #127, opened 2023-11-29
GKE DWS examples

mwielgus:


ahg-g:
can we please add a README file for this folder? what is DWS?

gke-dws-examples/manifests.yaml
```diff
@@ -0,0 +1,11687 @@
+apiVersion: v1
```
ahg-g, line 1: This installs a bunch of Kueue CRDs, are you sure you want this here?

alculquicondor: @mwielgus was is different here from the regular installation?
Is it just the feature gate?

Could we replace this with a script that:
1. downloads the installation manifest
2. adds the feature gate
3. `apply --server-side`

gke-dws-examples/job.yaml
```diff
@@ -0,0 +1,31 @@
...
+      containers:
+      - name: dummy-job
+        image: gcr.io/k8s-staging-perf-tests/sleep:v0.0.3
+        args: ["120s"]
+        resources:
+          requests:
+            cpu: "100m"
+            memory: "100Mi"
+            nvidia.com/gpu: 1
+          limits:
```
alculquicondor, line 29: Either copy everything or just set limits (as they are implicitly copied to requests).

gke-dws-examples/manifests.yaml
mwielgus: The problem with the kueue manifest is that it is 10k+ lines long. We don't want clients to edit this file as a starter.

mwielgus: Users may be reluctant to use scripts that do some magic they cannot check/control. Manifests, white big and cumbersome, seem simpler and easier to adopt.

ahg-g (approved):

PR #128, opened 2023-11-29
gke-platform: add default cluster labels to GKE clusters

andrewsykim:
Add default cluster labels so users can easily find clusters created using terraform modules in this repo

### DIFF
--- a/gke-platform/modules/gke_autopilot/variables.tf
+++ b/gke-platform/modules/gke_autopilot/variables.tf
@@ -30,6 +30,14 @@ variable "cluster_name" {
   default     = "ml-cluster"
 }
 
+variable "cluster_labels" {
+  type        = map
+  description = "GKE cluster labels"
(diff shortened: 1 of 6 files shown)


gke-platform/modules/gke_autopilot/variables.tf
```diff
@@ -30,6 +30,14 @@ variable "cluster_name" {
   default     = "ml-cluster"
 }
 
+variable "cluster_labels" {
+  type        = map
+  description = "GKE cluster labels"
+  default     =  {
+    created-by = "ai-on-gke"
```
andrewsykim, line 37: Initially had this as:
```
created-by: github.com/GoogleCloudPlatform/ai-ongke
```

But lable values can only be lower case and special characters like `.` and `/` are not allowed.

imreddy13 (approved):

PR #140, opened 2023-12-14
TPU Multi-Host Validating Webhook

ryanaoleary:
Add a validating webhook to enforce certain requirements when deploying TPU multi-host pod slices. Since the topology of the pod slice is statically known, we should enforce that the number of workers in a TPU worker group matches the number of hosts in the pod slice. Additionally, TPU workers in a group should be scheduled with affinity on the same node pool.


danielvegamyhre:
@ryanaoleary Can you add some context to the PR description? What are we trying to achieve with these changes?

ryanaoleary:
> e scheduled with affinity on the same pod slice.

Done.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -123,6 +124,84 @@ func injectHostnames(hostNames string, workerGroupSpec ray.WorkerGroupSpec, work
...
+		}
+		workerGroupAffinity, err := checkWorkerGroupAffinity(workerGroupSpec)
+		if err != nil {
+			return nil, err
+		}
+
+		if !(workersMatchTopology && workerGroupAffinity) {
+			admit = false
+			status = "Failure"
+			if !workersMatchTopology {
```
danielvegamyhre, line 184: If both of these errors can occur at the same time (workers don't match topology, worker group affinity set incorrectly, should the error message show both of these, instead of just one?

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -302,6 +381,31 @@ func main() {
...
+				return
+			}
+			admissionReview.Response = response
+			responseBytes, err := json.Marshal(admissionReview)
+			if err != nil {
+				klog.Errorf("Failed to encode response: %s", err)
+				return
+			}
+			fmt.Fprint(w, string(responseBytes))
+			return
```
danielvegamyhre, line 405: nit: do we need this return here?

ryanaoleary: Done.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -49,28 +50,28 @@ func containerRequestingTPUs(containers ...corev1.Container) bool {
...
 }
 
-// check if request is for TPU multi-host
-func isTPUMultiHost(topology string) (bool, error) {
+func getNumTPUHosts(topology string) (int, error) {
 	topologyVals :=  strings.Split(topology, "x")
-	values := make([]int, len(topologyVals))
-	for i := 0; i < len(values); i++ {
+	vms := 1
+	for i := 0; i < len(topologyVals); i++ {
```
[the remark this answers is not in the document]
ryanaoleary: Done.

ray-on-gke/kuberay-tpu-webhook/main.go
ryanaoleary: Done.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -123,6 +124,84 @@ func injectHostnames(hostNames string, workerGroupSpec ray.WorkerGroupSpec, work
...
+		}
+
+		if hosts != numWorkers {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func checkWorkerGroupAffinity(workerGroupSpec ray.WorkerGroupSpec) (bool, error) {
```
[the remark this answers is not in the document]
ryanaoleary: Done.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -123,6 +124,84 @@ func injectHostnames(hostNames string, workerGroupSpec ray.WorkerGroupSpec, work
 	}
 }
 
+func checkWorkersMatchTopology(workerGroupSpec ray.WorkerGroupSpec) (bool, error) {
+	numWorkers := int(*workerGroupSpec.Replicas)
+	if containerRequestingTPUs(workerGroupSpec.Template.Spec.Containers...) {
+		topology := workerGroupSpec.Template.Spec.NodeSelector["cloud.google.com/gke-tpu-topology"]
```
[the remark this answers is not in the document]
ryanaoleary: We expect the user to set it themselves so it's not guaranteed, I'll add a check if it exists or not.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -123,6 +124,84 @@ func injectHostnames(hostNames string, workerGroupSpec ray.WorkerGroupSpec, work
 	}
 }
 
+func checkWorkersMatchTopology(workerGroupSpec ray.WorkerGroupSpec) (bool, error) {
+	numWorkers := int(*workerGroupSpec.Replicas)
```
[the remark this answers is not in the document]
ryanaoleary: workerGoupSpec.Replicas should always be set, but I added in a check to see if it's nil and return an error if so.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -123,6 +124,84 @@ func injectHostnames(hostNames string, workerGroupSpec ray.WorkerGroupSpec, work
...
+	if containerRequestingTPUs(workerGroupSpec.Template.Spec.Containers...) {
```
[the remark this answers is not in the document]
ryanaoleary: I believe for it to be a valid workergroup, all these fields should exist. I'm not sure if Ray enforces that they're included though so I'll add checks to ensure they're not nil.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -89,7 +90,11 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
 }
 
 func genDNSHostnames(workerGroupSpec ray.WorkerGroupSpec) string {
-	numWorkers := int(*workerGroupSpec.Replicas)
+	replicas := workerGroupSpec.Replicas
+	if replicas == nil {
+		klog.Error("workerGroupSpec replicas not set")
```
danielvegamyhre, line 95: `klog.Error()` just logs with ERROR severity level and continues the program execution, so if replicas is nil you'll get a nil dereference panic in the subsequent lines.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -123,25 +132,138 @@ func injectHostnames(hostNames string, workerGroupSpec ray.WorkerGroupSpec, work
...
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func isScheduledWithAffinity(workerGroupSpec ray.WorkerGroupSpec) (bool, error) {
+	containers := workerGroupSpec.Template.Spec.Containers
+	if containers == nil {
+		klog.Error("Container path not specified")
```
danielvegamyhre, line 165: What criteria are you using to determine when to simply log an error and continue execution, versus return the error? I see both throughout this PR so it is unclear. For many of these errors it seems like you'd want the current execution path to halt and return an error, but for the webhook to keep running, so you'd probably want to return the error.

ray-on-gke/kuberay-tpu-webhook/main.go
ryanaoleary: That makes sense, I changed it here to return an error if replicas isn't set

ray-on-gke/kuberay-tpu-webhook/main.go
ryanaoleary: I went through and changed most of the `klog.Error` instances to return an informative error instead, I had been using klog where it seemed like the logic wouldn't continue anyway.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -88,74 +93,192 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
...
+			subdomainPatch, hostNamesPatch := patch{"op": "add",}, patch{"op": "add",}
+			subdomainPath := fmt.Sprintf("/spec/workerGroupSpecs/%d/template/spec/subdomain", workerGroupIndex)
+			envPath := fmt.Sprintf("/spec/workerGroupSpecs/%d/template/spec/containers/%d/env", workerGroupIndex, j)
 			tpuWorkerHostNames := corev1.EnvVar{
 				Name:  "TPU_WORKER_HOSTNAMES",
 				Value: hostNames,
 			}
+			subdomainPatch["path"] = subdomainPath
+			subdomainPatch["value"] = "tpu-worker-group-svc"
 			if len(container.Env) == 0 {
```
danielvegamyhre, line 128: Can you add a comment above this conditional logic explaining why the logic is different if the container has Env values already or not?

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -88,74 +93,192 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
...
+	if workerGroupSpecs == nil {
+		return nil, errors.New("WorkerGroupSpecs not specified")
+	}
+	for i := 0; i < len(workerGroupSpecs); i++ {
+		workerGroupSpec := workerGroupSpecs[i]
+		workersMatchTopology, err := checkWorkersMatchTopology(workerGroupSpec)
+		if err != nil {
+			return nil, err
+		}
+		workerGroupAffinity, err := isScheduledWithAffinity(workerGroupSpec)
```
danielvegamyhre, line 208: `workerGroupAffinity` sounds like an object name, not a boolean name, maybe it change it something like `hasCorrectAffinity` or something.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -88,74 +93,192 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
...
-			path := fmt.Sprintf("/spec/workerGroupSpecs/%d/template/spec/containers/%d/env", workerGroupIndex, j)
+			subdomainPatch, hostNamesPatch := patch{"op": "add",}, patch{"op": "add",}
+			subdomainPath := fmt.Sprintf("/spec/workerGroupSpecs/%d/template/spec/subdomain", workerGroupIndex)
+			envPath := fmt.Sprintf("/spec/workerGroupSpecs/%d/template/spec/containers/%d/env", workerGroupIndex, j)
 			tpuWorkerHostNames := corev1.EnvVar{
 				Name:  "TPU_WORKER_HOSTNAMES",
 				Value: hostNames,
 			}
+			subdomainPatch["path"] = subdomainPath
+			subdomainPatch["value"] = "tpu-worker-group-svc"
```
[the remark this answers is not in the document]
richardsliu: Better yet, define this as a constant so it's easier to find/replace if needed.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -88,74 +93,192 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
...
+	}
+	numWorkers := int(*replicas)
+	containers := workerGroupSpec.Template.Spec.Containers
+	if containers == nil {
+		return false, errors.New("Container path not specified")
+	}
+	if containerRequestingTPUs(containers...) {
+		topology := workerGroupSpec.Template.Spec.NodeSelector["cloud.google.com/gke-tpu-topology"]
+		if topology == "" {
+			klog.Error("TPU topology not specified")
```
richardsliu, line 153: Return err here?

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -88,74 +93,192 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
...
+		workerGroupAffinity, err := isScheduledWithAffinity(workerGroupSpec)
+		if err != nil {
+			return nil, err
+		}
+
+		if !(workersMatchTopology && workerGroupAffinity) {
+			admit = false
+			status = "Failure"
+			if !workersMatchTopology {
+				message = "Number of workers in worker group not equal to requested TPU hosts"
```
richardsliu, line 217: not equal to specified topology?

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -88,74 +93,192 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
...
+			} else if !workerGroupAffinity {
+				message = "TPU worker group requested without gke-placement-group nodeSelector"
+			} else {
+				message = "Missing gke-placement-group nodeSelector and worker not equal to TPU hosts"
```
richardsliu, line 221: Same

ray-on-gke/kuberay-tpu-webhook/main.go
ryanaoleary: Done.

ray-on-gke/kuberay-tpu-webhook/main.go
ryanaoleary: Done

ray-on-gke/kuberay-tpu-webhook/main.go
ryanaoleary: Returning an error here actually breaks the webhook - I believe it's because the tpu-topology may not be set yet and we want the webhook to keep running rather than exit out.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -199,15 +322,29 @@ func mutatePod(admissionReview *admissionv1.AdmissionReview) (*admissionv1.Admis
...
+	if topology == "" {
+		klog.Error("TPU topology not specified")
+	}
 	isMultiHost, _ := isTPUMultiHost(topology) // ignore error here because topology may not be set yet
-
+	containers := pod.Spec.Containers
+	if containers == nil {
+		return nil, errors.New("Container path not specified")
+	}
 	// inject the TPU_WORKER_ID environment variable into the container requesting TPUs
```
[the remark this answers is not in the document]
ryanaoleary: Done.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -88,74 +93,192 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
...
 func injectHostnames(hostNames string, workerGroupSpec ray.WorkerGroupSpec, workerGroupIndex int, patches *[]patch) {
-	for j := 0; j < len(workerGroupSpec.Template.Spec.Containers); j++ {
-		container := workerGroupSpec.Template.Spec.Containers[j]
+	containers := workerGroupSpec.Template.Spec.Containers
+	if containers == nil {
+		klog.Fatalf("Container path not specified")
+	}
+	for j := 0; j < len(containers); j++ {
+		container := containers[j]
 		if containerRequestingTPUs(container) {
```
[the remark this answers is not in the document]
ryanaoleary: Done.

ray-on-gke/kuberay-tpu-webhook/main.go
ryanaoleary: Done.

ray-on-gke/kuberay-tpu-webhook/main.go
ryanaoleary: Done.

ray-on-gke/kuberay-tpu-webhook/main.go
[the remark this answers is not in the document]
ryanaoleary: Done.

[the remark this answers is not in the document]
richardsliu: Actually thinking more on this - the service name should be dependent on the Kuberay cluster name since it's 1:1 with the cluster. So it should be something like "{kuberay-cluster-name}-tpu-worker-svc". 

The service can be created manually for now in the Terraform templates, but the longer term plan should be to create/delete it along with the RayCluster in the operator.

richardsliu (approved):

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -24,9 +26,10 @@ type slice struct {
 type patch map[string]any
 
 var (
-	tpuResourceName = corev1.ResourceName("google.com/tpu")
 	certPath = "/etc/kuberay-tpu-webhook/tls/tls.crt"
 	keyPath = "/etc/kuberay-tpu-webhook/tls/tls.key"
+	tpuResourceName = corev1.ResourceName("google.com/tpu")
+	headlessServiceName = "tpu-worker-svc"
```
richardsliu, line 32: Looks like you are extracting the Ray Cluster name in line 248, so can we base this name from that? 

That way if a user has multiple Ray clusters, each one can have a unique tpu-worker-svc based on the cluster name, and the same webhook can be reused.

ryanaoleary: Done.

PR #153, opened 2024-01-05
Cleanup the dws job example

mimowo:
I suggest to make the example minimal to reduce the chance of issues in case the node-pool name does not match the label. The node selector is not needed for the example to work, because Cluster Autoscaler (CA) will pick a node pool created with `--enable-queued-provisioning`.

In case there are multiple node-pools, per machine type, the preferred way of choosing the node pool would be using  `cloud.google.com/gke-accelerator` label configured at the Kueue's ResourceFlavor level.

In the current version, the yaml files here assume the DWS node-pool is named consistently, but the node-pool is created based on the docs in https://cloud.google.com/kubernetes-engine/docs/how-to/provisioningrequest, and users of the tutorial may not be aware of the connection, renaming the node-pool (or changing the accelerator type for the node-pool if we switch to  `cloud.google.com/gke-accelerator`).

### DIFF
--- a/gke-dws-examples/job.yaml
+++ b/gke-dws-examples/job.yaml
@@ -11,8 +11,6 @@ spec:
   suspend: true
   template:
     spec:
-      nodeSelector:
-        cloud.google.com/gke-nodepool: dws-nodepool
       tolerations:
       - key: "nvidia.com/gpu"
         operator: "Exists"


mimowo:
cc @kisieland @mwielgus @alculquicondor

mwielgus:
The node selector however sets a good example how to do it in more advanced/real environment. Without node selector, job may land on any node-pool, not necessairly the one from dws. So in order for the example to work, you need to modify resource falvor as weel.

mimowo:
> The node selector however sets a good example how to do it in more advanced/real environment. 

Maybe the real-world example with multiple nodepools (or machine types) fits better the `gke-batch-refarch` folder (but currently it does not mention DWS)?

In the current version, the yaml files here assume the DWS node-pool is named consistently, but the node-pool is created based on the docs in https://cloud.google.com/kubernetes-engine/docs/how-to/provisioningrequest, and users of the tutorial may not be aware of the connection, renaming the node-pool (or changing the accelerator type for the node-pool if we switch to  `cloud.google.com/gke-accelerator`). 

> Without node selector, job may land on any node-pool, not necessairly the one from dws. So in order for the example to work, you need to modify resource falvor as weel.

IIUC CA will choose a node-pool only from the DWS ones (those with `--enable-queued-provisioning`), so I think this example will work as long as the user has at least one DWS node pool. cc @kisieland to keep me honest here, but I've tested this on my cluster and it bound the DWS nodepool even though the default nodepool was also present.

mimowo:
/close 
In favor of https://github.com/GoogleCloudPlatform/ai-on-gke/pull/597

PR #154, opened 2024-01-05
Allow customizing job name.

canliu-aha:
The job names are used to provision intermediate resources (VMs, disks). When workflows are run concurrently, there would be name collisions. This change allows users to run more than 1 workflow at a time.

b/315384952

TESTED: manually

### DIFF
--- a/gke-disk-image-builder/cli/main.go
+++ b/gke-disk-image-builder/cli/main.go
@@ -41,6 +41,7 @@ func main() {
 	var containerImages stringSlice
 	projectName := flag.String("project-name", "", "name of a gcp project where the script will be run")
 	imageName := flag.String("image-name", "", "name of the image that will be generated")
+	jobName := flag.String("job-name", "secondary-disk-image", "name of the workflow job; no more than 50 characters")
 	zone := flag.String("zone", "", "zone where the resources will be used to create the image creator resources")
 	gcsPath := flag.String("gcs-path", "", "gcs location to dump the logs")
 	machineType := flag.String("machine-type", "n2-standard-16", "GCE instance machine type to generate the disk image")
@@ -61,6 +62,10 @@ func main() {
 		log.Panicf("invalid argument, timeout: %v, err: %v", timeout, err)
 	}
 
+	if len(*jobName) > 50 {
+		log.Panicf("invalid argument, job-name: %v cannot be longer than 50 characters, got: %v", *jobName, len(*jobName))
+	}
+
 	var auth builder.ImagePullAuthMechanism
 	switch *imagePullAuth {
 	case "":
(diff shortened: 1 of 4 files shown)


gke-disk-image-builder/cli/main.go
```diff
@@ -61,6 +62,10 @@ func main() {
 		log.Panicf("invalid argument, timeout: %v, err: %v", timeout, err)
 	}
 
+	if len(*jobName) > 50 {
+		log.Panicf("invalid argument, job-name: %v cannot be longer than 50 characters, got: %v", *jobName, len(*jobName))
```
ruiwen-zhao, line 66: just curious, is this limit from daisy?

ruiwen-zhao (approved):

gke-disk-image-builder/cli/main.go
canliu-aha: This is the limit introduced by GCP resource naming convention. We are using "${jobName}-bootable-disk" as one of the disk names and that one can't go over 63 characters, so jobName should be < 50 (counting again I think I should check for if `len(*jobName) >= 50` -- I'll update)

gke-disk-image-builder/imager.go
```diff
@@ -179,8 +182,8 @@ func GenerateDiskImage(ctx context.Context, req Request) error {
 		"detach-disk": {
 			DetachDisks: &daisy.DetachDisks{
 				&daisy.DetachDisk{
-					Instance:   fmt.Sprintf("%s-instance", name),
-					DeviceName: fmt.Sprintf("%s-disk", name),
+					Instance:   fmt.Sprintf("%s-instance", req.JobName),
+					DeviceName: deviceName,
```
BenjaminKazemi, line 186: `DeviceName: fmt.Sprintf("%s-disk", req.JobName)`

gke-disk-image-builder/imager.go
```diff
@@ -29,7 +29,8 @@ import (
 )
 
 const (
-	name = "secondary-disk-image"
+	deviceName = "secondary-disk-image-disk"
```
BenjaminKazemi, line 32: `deviceName` cannot be a constant as it depends on **JobName** (left a related comment below)

BenjaminKazemi (commented):
Please re-test the commit after addressing the comments. Thanks a lot for your contribution.

gke-disk-image-builder/imager.go
canliu-aha: Not sure if this is the correct understanding, but it reads like this is just the name to reference the attached disk on the VM, and it can be a constant unrelated to JobName? I hardcoded it to "secondary-disk-image-disk" because the `startup.sh` script uses this name.

image

BenjaminKazemi: Alright, we're good to go I think.
Turned out the device name a constant when you attach the disk to the instance.

BenjaminKazemi (approved):

PR #164, opened 2024-01-10
Support customizing image family name.

canliu-aha:
b/315387164

TESTED: manually run with --image-family-name flag.
Screenshot 2024-01-09 at 5 56 26 PM

### DIFF
--- a/gke-disk-image-builder/cli/main.go
+++ b/gke-disk-image-builder/cli/main.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"strings"
 	"time"
+	"regexp"
 
 	builder "github.com/GoogleCloudPlatform/ai-on-gke/gke-disk-image-builder"
 )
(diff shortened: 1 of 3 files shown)


gke-disk-image-builder/cli/main.go
```diff
@@ -37,10 +38,15 @@ func (s *stringSlice) Set(value string) error {
...
 
+var (
+	gcpResourceNameRegex = regexp.MustCompile("^[a-z]([-a-z0-9]*[a-z0-9])?$")
+)
+
 func main() {
 	var containerImages stringSlice
 	projectName := flag.String("project-name", "", "name of a gcp project where the script will be run")
 	imageName := flag.String("image-name", "", "name of the image that will be generated")
+	imageFamilyName := flag.String("image-family-name", "secondary-disk-image", "name of the image family associated with the created disk image")
```
danielvegamyhre, line 49: Is "secondary-disk-image" a valid image family name? I haven't heard of it, if it's not let's not set it as a default.

gke-disk-image-builder/cli/main.go
```diff
@@ -37,10 +38,15 @@ func (s *stringSlice) Set(value string) error {
 	return nil
 }
 
+var (
+	gcpResourceNameRegex = regexp.MustCompile("^[a-z]([-a-z0-9]*[a-z0-9])?$")
```
danielvegamyhre, line 42: Where did you find the rules for GCP resource names?

canliu-aha: https://cloud.google.com/compute/docs/naming-resources

The regex is here: https://screenshot.googleplex.com/5H4GFNpxnDabQJX

gke-disk-image-builder/cli/main.go
canliu-aha: It is the image family name currently hardcoded in the disk image builder code, so I'm using it as the default. Shall I change to something else?

danielvegamyhre: Oh weird, I never noticed that, ok sounds good.

danielvegamyhre (approved):

danielvegamyhre:
@canliu-aha Before I merge this, can you confirm you have manually tested these changes?

canliu-aha:
Yes, I confirmed I've manually tested the behavior.

PR #185, opened 2024-02-05
Cleaned up saxml http server container image

arueth:
Reduced the image from ~2GB to 126.8MB

Screenshot 2024-02-05 2 28 09 PM


alizaidis (approved):
LGTM

PR #197, opened 2024-02-14
Ray TPU Webhook RayJob Support

ryanaoleary:
Extends the mutating admission webhook for multi-host TPUs with Kuberay to support RayJob creation requests. This PR causes the webhook to intercept RayJob creation requests with the `app.kubernetes.io/name: kuberay` label and inject the `subdomain` field and `TPU_WORKER_HOSTNAMES` environment variable. This PR has been manually tested using `kubectl describe` to verify correct variable injection.


ryanaoleary:
In order for the webhook to work with RayJobs, it's necessary to annotate the RayJob with the label `app.kubernetes.io/name: kuberay`, since the mutating webhook configuration uses that label as a selector. I wanted to check whether this is reasonable, or if it'd be preferable to remove the label selectors entirely and have the webhook intercept all RayClusters, RayJobs, and Pods created while it's running.

ray-on-gke/kuberay-tpu-webhook/main.go
```diff
@@ -107,6 +107,20 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
 	return &rayCluster, nil
 }
 
+// unmarshal rayjob from admission request
+func extractRayJob(admissionReview *admissionv1.AdmissionReview) (*ray.RayJob, error) {
```
richardsliu, line 111: Do we have a test for this?

ryanaoleary: Currently just manual tests - using `kubectl describe` on the RayJob to verify env vars are injected and viewing the webhook logs.

imreddy13:
@ryanaoleary we did a refactor in https://github.com/GoogleCloudPlatform/ai-on-gke/commit/6aeff71532b7210e1bc50f913abf8bdb5c05427c, please move your changes under https://github.com/GoogleCloudPlatform/ai-on-gke/tree/main/applications/ray

PR #200, opened 2024-02-14
Makefile and README Changes

ryanaoleary:
This PR removes unnecessary `docker-build` and `docker-push` instructions from the README (the image is now managed through a build pipeline) and removes unnecessary `go build` statements from several `Make` commands. This PR also moves the TPU webhook under modules and updates the webhook README and `applications/ray` TPU_GUIDE with updated directory paths and instructions given the `ai-on-gke` refactoring.


modules/kuberay-tpu-webhook/README.md
imreddy13, line 1: /modules is meant to be for TF modules (so each "module" is independently usable/installable by tf apply) - I don't think that applies to the TPU webhook. Can we make it part of modules/kuberay-operator or modules/kuberay-cluster instead? And have a var flag to install it or not (there's already an enable_tpu flag, maybe reuse it)? wdyt @richardsliu ?

ryanaoleary: I think that sounds good, I'll move it to modules/kuberay-operator and make it part of the terraform install when `enable_tpu = true` in a follow-up PR. I'll leave it in applications/ray for now since this PR was focused on fixing some README and Makefile issues.

richardsliu (approved):

PR #221, opened 2024-02-22
Improve IAP setup process

blackzlq:
Design doc: https://docs.google.com/document/d/1aubk94Pd6qt1G2B2e-awOQ2ranzwgnSLikj5I6fBUwY/edit?resourcekey=0-2d-BsBiGOkwWrGb5aDYDTg&tab=t.0

1. Change the authenticator to auto fetch the backend service
2. add related permissions
3. changed permission and allow list to type list(string)

Tested-by: zlq

Tested:
Sandbox

### DIFF
--- a/applications/rag/main.tf
+++ b/applications/rag/main.tf
@@ -97,8 +97,8 @@ module "jupyterhub" {
   gcs_bucket  = var.gcs_bucket
   add_auth    = false # TODO: Replace with IAP.
 
-  gcp_service_account    = var.jupyter_service_account
-  create_service_account = var.create_jupyter_service_account
+  gcp_and_k8s_service_account    = var.jupyter_service_account
+  create_service_account         = var.create_jupyter_service_account
 
   # IAP Auth parameters
   brand                   = var.brand
--- a/applications/rag/workloads.tfvars
+++ b/applications/rag/workloads.tfvars
@@ -52,4 +52,4 @@ url_domain_addr   = ""
 url_domain_name   = ""
 client_id         = ""
 client_secret     = ""
-members_allowlist = "allAuthenticatedUsers,user:<email>"
+members_allowlist = ["allAuthenticatedUsers", "user:<email>"]
--- a/modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
+++ b/modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
@@ -22,14 +22,50 @@ from traitlets import Unicode
 from urllib import parse
 from google.auth.transport import requests
 from google.oauth2 import id_token
+from googleapiclient import discovery
+import google.auth
+
+
+def list_backend_services_ids(project_id, keyword):
+    credentials, _ = google.auth.default()
+    service = discovery.build('compute', 'v1', credentials=credentials)
+    request = service.backendServices().list(project=project_id)
+    response = request.execute()
+
+    filtered_service_ids = [
+        service['id'] for service in response.get('items', [])
+        if keyword.lower() in service['name'].lower()
+    ]
+
+    return filtered_service_ids
+
 
 class IAPUserLoginHandler(BaseHandler):
     def get(self):
         header_name = self.authenticator.header_name
-        
+
         auth_header_content = self.request.headers.get(header_name, "") if header_name else None
-        expected_audience = self.authenticator.expected_audience
-    
+
+        # Extract project ID, namespace, and backend config name from the authenticator
+        project_id = self.authenticator.project_id
+        namespace = self.authenticator.namespace
+        service_name = self.authenticator.service_name
+        print("Project ID:", project_id)
+        print("Namespace:", namespace)
+        print("Backend Config Name:", service_name)
+
+        # Construct the keyword from namespace and backend config name
+        keyword = namespace + "-" + service_name
+        print("Keyword:", keyword)
+
+        # List GCP backend services IDs based on the project ID and keyword
+        gcp_backend_services_ids = list_backend_services_ids(project_id, keyword)
+        print("GCP Backend Services IDs:", gcp_backend_services_ids)
+
+        # Construct expected audiences from the GCP backend services IDs
+        expected_audiences = [f"/projects/{self.authenticator.project_number}/global/backendServices/{service_id}" for service_id in gcp_backend_services_ids]
+        print("Expected Audiences:", expected_audiences)
+
         if self.authenticator.header_name != "X-Goog-IAP-JWT-Assertion":
             raise web.HTTPError(400, 'X-Goog-IAP-JWT-Assertion is the only accepted Header')
         elif bool(auth_header_content) == 0:
@@ -37,19 +73,19 @@ class IAPUserLoginHandler(BaseHandler):
         else:
             _, user_email, err = validate_iap_jwt(
                 auth_header_content,
-                expected_audience
+                expected_audiences
             )
             if err:
                 raise Exception(f'Ran into error: {err}')
             else:
                 logging.info(f'Successfully validated!')
-        
+
         username = user_email.lower().split("@")[0]
         user = self.user_from_username(username)
 
         self.set_login_cookie(user)
         self.redirect(url_path_join(self.hub.server.base_url, 'home'))
-        
+
 class GCPIAPAuthenticator(Authenticator):
     """
     Accept the authenticated JSON Web Token from IAP Login.
@@ -58,8 +94,8 @@ class GCPIAPAuthenticator(Authenticator):
     """
     header_name = Unicode(
         config=True,
-        help="""HYYP header to inspect for the authenticated JWT.""")
-    
+        help="""HTTP/HTTPS header to inspect for the authenticated JWT.""")
+
     cookie_name = Unicode(
         config=True,
         help="""The name of the cookie field used to specify the JWT token""")
@@ -67,12 +103,27 @@ class GCPIAPAuthenticator(Authenticator):
     param_name = Unicode(
         config=True,
         help="""The name of the query parameter used to specify the JWT token""")
-    
-    expected_audience = Unicode(
+
+    project_id = Unicode(
         default_value='',
         config=True,
-        help="""Expected Audience of the authenication JWT""")
-    
+        help="""Expected project_id""")
+
+    project_number = Unicode(
+        default_value='',
+        config=True,
+        help="""Expected project_number""")
+
+    namespace = Unicode(
+        default_value='',
+        config=True,
+        help="""Expected namespace""")
+
+    service_name = Unicode(
+        default_value='',
+        config=True,
+        help="""Expected backend_config_name""")
+
     secret = Unicode(
         config=True,
         help="""Shared secret key for signing JWT token""")
@@ -80,12 +131,12 @@ class GCPIAPAuthenticator(Authenticator):
     def get_handlers(self, app):
         return [(r'login', IAPUserLoginHandler)]
 
-def validate_iap_jwt(iap_jwt, expected_audience):
+def validate_iap_jwt(iap_jwt, expected_audiences):
     """Validate an IAP JWT.
 
     Args:
       iap_jwt: The contents of the X-Goog-IAP-JWT-Assertion header.
-      expected_audience: The Signed Header JWT audience. See
+      expected_audiences: The Signed Header JWT audiences. See
           https://cloud.google.com/iap/docs/signed-headers-howto
           for details on how to get this value.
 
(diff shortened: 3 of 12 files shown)


modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
```diff
@@ -59,33 +95,48 @@ class GCPIAPAuthenticator(Authenticator):
...
         config=True,
         help="""The name of the cookie field used to specify the JWT token""")
 
     param_name = Unicode(
         config=True,
         help="""The name of the query parameter used to specify the JWT token""")
-    
-    expected_audience = Unicode(
+
+    project_id = Unicode(
```
chiayi, line 107: Is there a reason we are adding project_id, project_number, and namespace? Will we need them and will the JWT tokens contain these fields?

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
```diff
@@ -59,33 +95,48 @@ class GCPIAPAuthenticator(Authenticator):
     header_name = Unicode(
         config=True,
         help="""HYYP header to inspect for the authenticated JWT.""")
```
chiayi, line 97: Could you do a quick fix here? It should be 'HTTP' instead of 'HYYP'

chiayi (commented):
Just some NITs and questions.

applications/rag/workloads.tfvars
```diff
@@ -43,7 +43,7 @@ dataset_embeddings_table_name = "googlemaps_reviews_db"
 
 ## IAP config
 add_auth                = false # Set to true when using auth with IAP
-brand                   = "projects/<prj-number>/brands/<prj-number>"
+brand                   = ""
```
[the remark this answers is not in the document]
blackzlq: Default is to create the brand, but I can add it back

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
```diff
@@ -22,34 +22,70 @@
...
-        expected_audience = self.authenticator.expected_audience
-    
+
+        # Extract project ID, namespace, and backend config name from the authenticator
+        project_id = self.authenticator.project_id
+        namespace = self.authenticator.namespace
+        service_name = self.authenticator.service_name
+        print("Project ID:", project_id)
+        print("Namespace:", namespace)
+        print("Backend Config Name:", service_name)
```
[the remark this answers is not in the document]
blackzlq: I added TODO, yes, we need to do that

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
blackzlq: LOL, just notice, for sure

applications/rag/main.tf
```diff
@@ -97,8 +97,8 @@ module "jupyterhub" {
   gcs_bucket  = var.gcs_bucket
   add_auth    = false # TODO: Replace with IAP.
 
-  gcp_service_account    = var.jupyter_service_account
-  create_service_account = var.create_jupyter_service_account
+  gcp_and_k8s_service_account    = var.jupyter_service_account
```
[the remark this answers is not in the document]
blackzlq: This is just how workload identity will work, but yeah, we can do that

modules/jupyter/main.tf
```diff
@@ -63,17 +63,32 @@ module "iap_auth" {
...
+  name                = var.gcp_and_k8s_service_account
   namespace           = var.namespace
   project_id          = var.project_id
-  roles               = ["roles/cloudsql.client", "roles/artifactregistry.reader", "roles/storage.admin"]
-  depends_on = [ module.iap_auth ]
+  roles               = concat(var.predefined_iam_roles, var.gcp_service_account_iam_roles)
+  depends_on          = [ module.iap_auth ]
+}
+
+resource "kubernetes_annotations" "hub" {
```
umeshkumhar, line 73: this is not required, jupyterhub-workload-identity will create gcp and k8s sa with same name and configure annotations.

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
blackzlq: expected_audiences = [f"/projects/{self.authenticator.project_number}/global/backendServices/{service_id}" for service_id in gcp_backend_services_ids]

modules/jupyter/main.tf
```diff
@@ -94,16 +109,26 @@ resource "helm_release" "jupyterhub" {
   values = [
     templatefile("${path.module}/jupyter_config/config-selfauth.yaml", {
       password            = var.add_auth ? "dummy" : random_password.generated_password[0].result
-      service_id          = var.add_auth ? (data.google_compute_backend_service.jupyter-ingress[0].generated_id != null ? data.google_compute_backend_service.jupyter-ingress[0].generated_id : "no-id-yet") : "no-id-yet"
-      backend_config      = var.service_name
+      project_id          = var.project_id
       project_number      = data.google_project.project.number
+
+      # Support legacy image.
+      service_id          = var.add_auth ? (data.google_compute_backend_service.jupyter-ingress[0].generated_id != null ? data.google_compute_backend_service.jupyter-ingress[0].generated_id : "no-id-yet") : "no-id-yet"
```
umeshkumhar, line 116: this is not required after this change.?

modules/jupyter/jupyter_config/config-selfauth.yaml
```diff
@@ -30,6 +30,11 @@ hub:
       authenticator_class: ${authenticator_class}
     GCPIAPAuthenticator: 
       header_name: 'X-Goog-IAP-JWT-Assertion'
+      project_id: '${project_id}'
+      namespace: '${namespace}'
+      service_name: '${service_name}'
+      project_number: '${project_number}'
+      # support legacy image
       expected_audience: '/projects/${project_number}/global/backendServices/${service_id}'
```
umeshkumhar, line 38: this will be generated automatically, we dont need to pass it?

modules/jupyter/main.tf
blackzlq: I assume the add auth is still optional?

modules/jupyter/jupyter_config/config-selfauth.yaml
blackzlq: support legacy, because I can not modify the repo the same time, after this change merged, I will update repo and remove this line

applications/rag/workloads.tfvars
```diff
@@ -42,7 +42,7 @@ jupyter_service_account = "jupyter-system-account"
 dataset_embeddings_table_name = "googlemaps_reviews_db"
 
 ## IAP config
-add_auth                = false # Set to true when using auth with IAP
+add_auth                = var.add_auth # Set to true when using auth with IAP
```
imreddy13, line 45: this should be true or false, not a var. this is workloads.tfvars where the user sets explicit values.

i would set it to false unless you verified it works E2E with RAG. we can set it to true in a followup PR.

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
```diff
@@ -22,34 +22,70 @@
...
+
         auth_header_content = self.request.headers.get(header_name, "") if header_name else None
-        expected_audience = self.authenticator.expected_audience
-    
+
+        # Extract project ID, namespace, and backend config name from the authenticator
+        project_id = self.authenticator.project_id
+        namespace = self.authenticator.namespace
+        service_name = self.authenticator.service_name
+        print("Project ID:", project_id)
```
imreddy13, line 53: delete these debug prints?

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
```diff
@@ -22,34 +22,70 @@
...
+        keyword = namespace + "-" + service_name
+        print("Keyword:", keyword)
+
+        # List GCP backend services IDs based on the project ID and keyword
+        gcp_backend_services_ids = list_backend_services_ids(project_id, keyword)
+        print("GCP Backend Services IDs:", gcp_backend_services_ids)
+
+        # Construct expected audiences from the GCP backend services IDs
+        expected_audiences = [f"/projects/{self.authenticator.project_number}/global/backendServices/{service_id}" for service_id in gcp_backend_services_ids]
+        print("Expected Audiences:", expected_audiences)
```
imreddy13, line 67: did you intend to leave these prints?

modules/jupyter/main.tf
```diff
@@ -63,17 +63,32 @@ module "iap_auth" {
 module "jupyterhub-workload-identity" {
   source              = "terraform-google-modules/kubernetes-engine/google//modules/workload-identity"
   use_existing_gcp_sa = !var.create_service_account
-  name                = var.gcp_service_account
+  name                = var.gcp_and_k8s_service_account
   namespace           = var.namespace
   project_id          = var.project_id
-  roles               = ["roles/cloudsql.client", "roles/artifactregistry.reader", "roles/storage.admin"]
-  depends_on = [ module.iap_auth ]
+  roles               = concat(var.predefined_iam_roles, var.gcp_service_account_iam_roles)
```
imreddy13, line 69: do we need both sets of roles? i think they overlap. can you verify?

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
```diff
@@ -22,34 +22,70 @@
...
+        print("Project ID:", project_id)
+        print("Namespace:", namespace)
+        print("Backend Config Name:", service_name)
+
+        # Construct the keyword from namespace and backend config name
+        keyword = namespace + "-" + service_name
+        print("Keyword:", keyword)
+
+        # List GCP backend services IDs based on the project ID and keyword
+        gcp_backend_services_ids = list_backend_services_ids(project_id, keyword)
```
imreddy13, line 62: shouldn't we do this in a loop until the fetch succeeds?

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
blackzlq: Yes, it can help user to understand

applications/rag/workloads.tfvars
blackzlq: I have tested E2E, we can modify it to true after I push the image to repo, We have a tricky thing is that I can not make this code change and image updated the same time. So you can see I have a legacy support to ensure it will not break anything. But for sure IAP will be the default setting for jupyter. I will add it in next PR

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
blackzlq: I think this log is helpful to understand how the authenticator works.

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
blackzlq: I don't think so, because even we do the loop, user will refresh the page. So we can simply rely on user trigger, one thing I am thinking is cache, once we found maybe we can cache it.

modules/jupyter/main.tf
blackzlq: The first array is the permission we define and the second is user can define more permission, for example I want to my jupyter hub can access to other cluster. Since jupyter notebook may have some new requirement. The reason why I want to have two arrays is because I don't want to user to touch the default permission.

applications/rag/workloads.tfvars
imreddy13: sorry, can we just explicitly set this to "false" instead?

imreddy13: and update to true in the next PR

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
imreddy13: for my own understanding: since the cert takes 10-20 mins to provision & the user cannot login until it's done, it's fine to not have a loop here

imreddy13 (approved):

applications/rag/workloads.tfvars
blackzlq: changed to false

modules/jupyter/authentication/authenticator/gcpiapjwtauthenticator/gcpiapjwtauthenticator.py
blackzlq: Synced offline, we agree on this

PR #222, opened 2024-02-23
Add Private IP Configuration to CloudSQL

sunnylovestiramisu:
Terraform apply and tested the jobs.

The ray job and frontend app needs to run in the same network as the private IP configuration.


applications/rag/frontend/container/main.py
imreddy13, line 1: I'm assuming you built and tested this new container with the frontend?

imreddy13: Don't you need to fix applications/rag/frontend/main.tf frontend deployment to use the new image? https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/rag/frontend/main.tf#L88

I can grant you access to publish the image. See https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/rag/frontend/container/README.md for how to publish

applications/rag/workloads.tfvars
```diff
@@ -18,6 +18,9 @@ project_id = "<project_id>"
 cluster_name     = "<cluster_name>"
 cluster_location = "us-central1"
 
+## CloudSQL variables
+private_network  = "ml-network"
```
imreddy13, line 22: who creates this? is it during cluster creation? @umeshkumhar is there a better way to do this

imreddy13: README needs update for how to use this variable.

modules/cloudsql/main.tf
```diff
@@ -12,6 +12,24 @@
...
+
+resource "google_compute_global_address" "private_ip_address" {
+  name          = "private-ip-address"
+  purpose       = "VPC_PEERING"
+  address_type  = "INTERNAL"
+  prefix_length = 16
+  network       = data.google_compute_network.compute_network.id
+}
+
+resource "google_service_networking_connection" "private_vpc_connection" {
```
umeshkumhar, line 27: we will move this out to infrastructure module.

applications/rag/workloads.tfvars
sunnylovestiramisu: This is in the infra, so whatever you create there you use for Ray, CloudSQL and the Frontend right now.

modules/cloudsql/main.tf
sunnylovestiramisu: Hmm i am not sure, this is for CloudSQL setup only, so it is better to bundle with CloudSQL.

umeshkumhar:
@sunnylovestiramisu 
https://github.com/terraform-google-modules/terraform-google-sql-db/blob/v19.0.0/examples/mysql-psc/main.tf

lets use cloudsql module. it has additional user support + psc networking...

modules/cloudsql/main.tf
sunnylovestiramisu: The infra set up does not have google_service_networking_connection setup though? The problem is cloudsql private IP requires this to exist.

pwschuurman:
Opened #398 to address this feature request

PR #226, opened 2024-02-23
Add option to enable/disable Grafana for Ray dashboard

umeshkumhar:
Add option to enable or disable grafana for the ray dashboard. Enabling requires anonymous access

### DIFF
--- a/applications/ray/variables.tf
+++ b/applications/ray/variables.tf
@@ -42,6 +42,12 @@ variable "ray_namespace" {
   default     = "myray"
 }
 
+variable "enable_grafana_on_ray_dashboard" {
+  type    = bool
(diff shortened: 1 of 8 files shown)


imreddy13:
Can you add a more meaningful description?
"Add option to enable or disable grafana for the ray dashboard. Enabling requires anonymous access."

applications/ray/variables.tf
```diff
@@ -42,6 +42,11 @@ variable "ray_namespace" {
   default     = "myray"
 }
 
+variable "enable_ray_dashboard" {
```
imreddy13, line 45: can we call it enable_grafana_on_ray_dashboard? it's not used anywhere else, right?

imreddy13 (approved):

PR #227, opened 2024-02-23
Rename k8s ingress and backend config name

blackzlq:
Fix the naming mess for IAP.
k8s ingress should be named as proxy-public
backend service for now name it as iap-config-default

Tested-by: zlq

Tested:
Sandbox

### DIFF
--- a/gke-disk-image-builder/cli/main.go
+++ b/gke-disk-image-builder/cli/main.go
@@ -60,6 +60,8 @@ func main() {
 	timeout := flag.String("timeout", "20m", "Default timout for each step, defaults to 20m")
 	network := flag.String("network", "default", "VPC network to be used by GCE resources used for disk image creation.")
 	subnet := flag.String("subnet", "default", "subnet to be used by GCE resources used for disk image creation.")
+	storeSnapshotCheckSum := flag.Bool("store-snapshot-checksum", true, "calculate and store checksums of every snapshot directory.")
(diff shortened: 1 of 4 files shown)


applications/jupyter/main.tf
```diff
@@ -81,15 +81,15 @@ module "jupyterhub" {
...
-  url_domain_addr         = var.url_domain_addr
-  url_domain_name         = var.url_domain_name
-  members_allowlist       = var.members_allowlist
-  depends_on              = [module.gcs]
+  add_auth                  = var.add_auth
+  brand                     = var.brand
+  support_email             = var.support_email
+  client_id                 = var.client_id
+  client_secret             = var.client_secret
+  k8s_ingress_name          = var.k8s_ingress_name
```
umeshkumhar, line 89: this is not exactly a ingress.. this is the ingress - default backend k8s_service.

this is configured as default traffic that will be sent to this service.. (proxy-public, i.e. of jupyter)

imreddy13: proxy-public is not an ingress, it is a k8s service. i think the previous name is correct

applications/jupyter/main.tf
```diff
@@ -81,15 +81,15 @@ module "jupyterhub" {
...
+  k8s_backend_config_name   = var.k8s_backend_config_name
```
imreddy13, line 90: this can be k8s_ingress

imreddy13:
not convinced why we need this change

PR #230, opened 2024-02-23
get cloudbuild presubmit passing again

andrewsykim:
We recently refactored a bunch of terraform modules but the cloudbuild configuration was not updated. This PR gets cloudbuild presubmits passing again

### DIFF
--- a/applications/jupyter/workloads-without-iap.example.tfvars
+++ b/applications/jupyter/workloads-without-iap.example.tfvars
@@ -35,8 +35,8 @@ gcp_and_k8s_service_account   = "jupyter-service-account"
 add_auth = false
 brand = ""
 support_email = ""
-k8s_ingress_name = ""
-k8s_backend_config_name = ""
+k8s_ingress_name          = "jupyter-ingress"
+k8s_backend_config_name   = "jupyter-iap-config"
 k8s_backend_service_name = "proxy-public"
 
 url_domain_addr = ""
--- a/applications/jupyter/workloads-without-iap.example.tfvars
+++ b/applications/jupyter/workloads-without-iap.example.tfvars
@@ -29,7 +29,7 @@ cluster_membership_id = "" # required only for private clusters, default: cluste
 namespace = "jupyter"
 gcs_bucket = "<gcs-bucket>"
 create_service_account        = true
-gcp_service_account           = "jupyter-service-account"
+gcp_and_k8s_service_account   = "jupyter-service-account"
 
 # Jupyterhub with IAP
 add_auth = false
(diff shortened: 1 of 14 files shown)


applications/ray/main.tf
```diff
@@ -109,6 +109,7 @@ module "kuberay-cluster" {
   namespace              = var.ray_namespace
   project_id             = var.project_id
   enable_tpu             = data.google_container_cluster.default.enable_tpu
+  enable_gpu             = var.enable_gpu
```
andrewsykim, line 112: Ray clusters for tests should only use CPUs, the module only supports picking between GPU and TPUs, this is added so we can disable both

imreddy13 (approved):

applications/jupyter/workloads-without-iap.example.tfvars
```diff
@@ -29,17 +29,17 @@ cluster_membership_id = "" # required only for private clusters, default: cluste
...
 create_service_account        = true
-gcp_service_account           = "jupyter-service-account"
+gcp_and_k8s_service_account   = "jupyter-service-account"
 
 # Jupyterhub with IAP
 add_auth = false
 brand = ""
 support_email = ""
-default_backend_service = ""
+default_backend_service = "proxy-public"
```
imreddy13, line 38: Good catch!

modules/kuberay-cluster/kuberay.tf
```diff
@@ -25,19 +25,26 @@ resource "helm_release" "ray-cluster" {
   namespace        = var.namespace
   create_namespace = var.create_namespace
   version          = "0.6.1"
-  values = var.enable_autopilot ? [templatefile("${path.module}/kuberay-autopilot-values.yaml", {
-    gcs_bucket          = var.gcs_bucket
-    k8s_service_account = var.google_service_account
-    grafana_host        = var.grafana_host
-    })] : (var.enable_tpu ? [templatefile("${path.module}/kuberay-tpu-values.yaml", {
+
+  values = [
```
imreddy13, line 29: I wonder if there is a better way to do this templating - this could get ugly  (already is ugly)

cloudbuild.yaml
```diff
@@ -127,21 +133,28 @@ steps:
...
-        -var=service_account=$_USER_NAME-$SHORT_SHA-system-account \
+        -var=cluster_name=ml-$SHORT_SHA-$_PR_NUMBER-cluster \
+        -var=cluster_location=$_REGION \
+        -var=ray_namespace=ml-$SHORT_SHA \
+        -var=gcp_service_account=ray-sa-$SHORT_SHA \
+        -var=gcs_bucket=gke-aieco-ray-$SHORT_SHA \
         -auto-approve -no-color
 
-        gcloud container clusters delete ray-$SHORT_SHA-$_PR_NUMBER-cluster --zone $_ZONE  --project $PROJECT_ID --quiet
+        gcloud container clusters delete ml-$SHORT_SHA-$_PR_NUMBER-cluster --location $_REGION  --project $PROJECT_ID --quiet
```
hsachdevah, line 157: We should delete the cluster using Terraform destroy from infrastructure/ as 'create gke cluster' step above may create additional resources.

andrewsykim: I see that it's a variable now so we can disable it and use terraform to delete resources, I'll update to use that

PR #231, opened 2024-02-23
Fix DLP filter

blackzlq:
Missing env and permission after we migrate to gke cluster. Tested-by: zlq

Tested:
Sandbox


umeshkumhar (approved):
lgtm

PR #237, opened 2024-02-26
Fix the naming mess for IAP.

blackzlq:
k8s ingress should be named as proxy-public
backend service for now name it as iap-config-default

Tested-by: zlq

Tested:
Sandbox

### DIFF
--- a/applications/jupyter/main.tf
+++ b/applications/jupyter/main.tf
@@ -81,15 +81,16 @@ module "jupyterhub" {
   gcs_bucket                    = var.gcs_bucket
 
   # IAP Auth parameters
-  add_auth                = var.add_auth
-  brand                   = var.brand
-  support_email           = var.support_email
-  client_id               = var.client_id
-  client_secret           = var.client_secret
-  default_backend_service = var.default_backend_service
-  service_name            = var.service_name
-  url_domain_addr         = var.url_domain_addr
-  url_domain_name         = var.url_domain_name
-  members_allowlist       = var.members_allowlist
-  depends_on              = [module.gcs]
+  add_auth                  = var.add_auth
+  brand                     = var.brand
+  support_email             = var.support_email
+  client_id                 = var.client_id
+  client_secret             = var.client_secret
+  k8s_ingress_name          = var.k8s_ingress_name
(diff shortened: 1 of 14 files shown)


applications/jupyter/main.tf
```diff
@@ -81,15 +81,15 @@ module "jupyterhub" {
   gcs_bucket                    = var.gcs_bucket
 
   # IAP Auth parameters
-  add_auth                = var.add_auth
-  brand                   = var.brand
-  support_email           = var.support_email
-  client_id               = var.client_id
-  client_secret           = var.client_secret
-  default_backend_service = var.default_backend_service
```
chiayi, line 89: I think calling it Jupyterhub service ingress name of some sort (relating it to the Jupyterhub service) would be clearer, since it will always be the name of the Jupyterhub load balancer service name. WDYT?

chiayi (approved):
Overall LGTM, just one small question

applications/jupyter/main.tf
chiayi: Discussed offline, the default backed service could be used in other cases.

applications/jupyter/workloads-without-iap.example.tfvars
```diff
@@ -35,8 +35,8 @@ gcp_service_account           = "jupyter-service-account"
 add_auth = false
 brand = ""
 support_email = ""
-default_backend_service = ""
-service_name = ""
+k8s_ingress_name = ""
+k8s_backend_config_name = ""
```
chiayi, line 40: NIT: have k8s_backend_service_name here to make everything consistent? Even though it won't need it.

modules/jupyter/variables.tf
```diff
@@ -66,10 +61,21 @@ variable "project_id" {
   description = "GCP project id"
 }
 
-variable "service_name" {
+variable "k8s_ingress_name" {
+  type    = string
```
chiayi, line 65: Description?

chiayi (approved):
Just NITs, overall LTGM

modules/jupyter/variables.tf
blackzlq: Done

PR #239, opened 2024-02-27
[TPU Provisioner] Enable auto-repair for TPU node pools

danielvegamyhre:
We had auto-repair disabled during testing, we need to re-enable it.

cc @nstogner


nstogner (approved):
LGTM

thebinaryone1 (approved):

PR #251, opened 2024-02-28
Misc doc fixes

rsgowman:
See changes for details. Some extra reasoning behind the changes:
* Prefix all GCS buckets with project_id since GCS buckets are globally namespaced.
* gcloud storage buckets add-iam-policy-binding" only accepts bucket URLs


rsgowman:
@annapendleton

PR #262, opened 2024-02-29
Add HPA configuration

rsgowman:
This allows the tgi workload to scale up based on CPU demand (and
eventually other metrics).

NB: CPU is a poor choice for this workload, but acts as a baseline that
we can use to evaluate other metrics.


benchmarks/inference-server/text-generation-inference/hpa-templates/hpa.cpu.yaml.tftpl
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: tgi
+  namespace: ${namespace}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: tgi
+  minReplicas: 1
+  maxReplicas: 5
```
annapendleton, line 12: Please make the maxReplicas a variable in tfvars

rsgowman: Done. I've also done the same for minReplicas.

For min, I've defaulted to 1 (which is reasonable). For max, I've defaulted to 5 (which is arbitrary - but at least controls costs if people blindly apply this.)

annapendleton (approved):

PR #264, opened 2024-02-29
Add support to RAG application to create a cluster for the user, or allow them to bring their own cluster

umeshkumhar:
- Create a cluster with appropriate nodepools/other config on behalf of the user
- Still allow bringing your own cluster
- Update marketplace blueprints with cluster creation options

### DIFF
--- a/applications/rag/variables.tf
+++ b/applications/rag/variables.tf
@@ -18,7 +18,8 @@ variable "project_id" {
 }
 
 variable "cluster_name" {
-  type = string
+  type    = string
+  default = "new-cluster"
 }
 
 variable "cluster_membership_id" {
@@ -28,7 +29,8 @@ variable "cluster_membership_id" {
 }
 
 variable "cluster_location" {
-  type = string
+  type    = string
+  default = "us-central1"
 }
 
 variable "kubernetes_namespace" {
@@ -79,8 +81,8 @@ variable "rag_service_account" {
 }
 
 variable "create_gcs_bucket" {
-  type = bool
-  default = false
+  type        = bool
+  default     = false
   description = "Enable flag to create gcs_bucket"
 }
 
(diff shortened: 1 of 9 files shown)


applications/rag/variables.tf
```diff
@@ -18,7 +18,8 @@ variable "project_id" {
 }
 
 variable "cluster_name" {
-  type = string
+  type    = string
+  default = "new-cluster"
```
imreddy13, line 22: why do we need this?

applications/rag/variables.tf
```diff
@@ -28,7 +29,8 @@ variable "cluster_membership_id" {
 }
 
 variable "cluster_location" {
-  type = string
+  type    = string
+  default = "us-central1"
```
imreddy13, line 33: same

imreddy13: if we have to default, can we pick a region that has more L4 capacity ?

applications/rag/variables.tf
```diff
@@ -79,8 +81,8 @@ variable "rag_service_account" {
 }
 
 variable "create_gcs_bucket" {
-  type = bool
-  default = false
+  type        = bool
+  default     = false
```
imreddy13, line 85: shouldnt it be default true? or jupyterhub/ray wont work

imreddy13 (dismissed):

applications/rag/variables.tf
umeshkumhar: default false is for marketplace..

imreddy13 (approved):

artemvmin:
FYI something in this commit changed `tf apply` (pre-deployment) plan generation from O(seconds) to O(minutes). Unless I made a mistake, we need to solve this or revert this change.

```
# at HEAD (this commit):
time (tf apply -var-file workloads.tfvars)` 
...
real    5m16.797s
user    8m49.858s
sys     0m22.037s
```
```
# at `HEAD~1`
time (tf apply -var-file workloads.tfvars)` 
...
real    0m20.667s
user    0m26.451s
sys     0m3.219s
```

PR #269, opened 2024-03-01
[TPU Provisioner] Bump node pool deletion check interval

danielvegamyhre:
This is needed to allow for more time for Jobs (and pods) to be recreated upon JobSet restart, since JobSet now uses foreground deletion to avoid [other issues](https://github.com/kubernetes-sigs/jobset/issues/392), so Job recreation is blocked until all of the pods from the previous Job are completely deleted. This means it may take longer than 4 seconds for the pods to begin rescheduling during failure recovery, so we should bump the interval to 60 seconds.


nstogner:
LGTM

thebinaryone1 (approved):

PR #263, opened 2024-02-29
Add ability to scale on tgi custom metrics

rsgowman:

PR #272, opened 2024-03-02
Gke triton benchmark

kaushikmitr:


kaushikmitr:
Added Terraform scripts for using Triton + TensorRTLLM on GKE.

benchmarks/inference-server/triton/manifest-templates/triton-tensorrtllm-inference-gs.tftpl
```diff
@@ -0,0 +1,76 @@
...
+      serviceAccountName: ${ksa}  # Replace ${ksa} with your Kubernetes Service Account name
+      volumes:
+        - name: dshm
+          emptyDir:
+            medium: Memory
+            sizeLimit: 1Gi
+        - name: all-models-volume
+          emptyDir: {}
+      initContainers:
+        - name: init-gcs-download
```
annapendleton, line 41: Note that this additional storage option still relies on copying the long-term stored object into ephemeral storage. We also will want to support reading directly from external storage in the future, here's a general overview of what is offered:
https://cloud.google.com/kubernetes-engine/docs/concepts/storage-overview

This automated yaml is a great starting point for integrating the GKE managed external storage options as well, as the main difference is how the external storage is mounted to the container, I would expect the rest of the triton startup would be the same.

benchmarks/inference-server/triton/README.md
```diff
@@ -0,0 +1,198 @@
...
+   gpu_count = X
+   ```
+   * If using method 2 In Step 1 above:
+   
+   ```bash
+   template_path = "path_to_manifest_template/triton-tensorrtllm-inference-gs.tftpl"
+   image_path = "nvcr.io/nvidia/tritonserver:23.10-trtllm-python-py3"
+   gpu_count = X
+   ```
+   and also update the `triton-tensorrtllm-inference-gs.tftpl` with the path to the gcs repo under InitContainer and the command to launch the container under Container section.
```
annapendleton, line 107: We should have a variable defined in the .tfvars file and pipe through the automation rather than asking customer to edit .tftpl files directly

benchmarks/inference-server/triton/variables.tf
```diff
@@ -0,0 +1,82 @@
...
+}
+
+variable "huggingface-secret" {
+  description = "name of the kubectl huggingface secret token"
+  type        = string
+  nullable    = true
+  default     = "huggingface-secret"
+}
+
+variable "template_path" {
```
annapendleton, line 77: We can automate this step by choosing which template for the customer to use based on if they provide the gs:// path or not.

See HPA template choice config as an example: 
https://github.com/GoogleCloudPlatform/ai-on-gke/pull/263/files#diff-c4d8a09078e87c6689697d9892fce3dcbbd7f89d19402db567c76e03389417df

benchmarks/inference-server/triton/manifest-templates/triton-tensorrtllm-inference-gs.tftpl
```diff
@@ -0,0 +1,76 @@
...
+          volumeMounts:
+            - name: all-models-volume
+              mountPath: /all_models
+      containers:
+        - name: triton-inference
+          image: "${image_path}"  # Replace ${image_path} with the actual image path
+          workingDir: /opt/tritonserver
+          #command: ["/bin/sleep", "3600"]
+          command: ["/bin/bash", "-c"]
+          args: ["pip install sentencepiece protobuf && huggingface-cli login --token $HUGGINGFACE_TOKEN && mpirun --allow-run-as-root -n 1 /opt/tritonserver/bin/tritonserver --model-repository=/all_models/inflight_batcher_llm --disable-auto-complete-config --backend-config=python,shm-region-prefix-name=prefix0_ :"]
```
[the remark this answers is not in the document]
annapendleton: Ah I missed this earlier - this still utilizes custom image path.

My understanding with tensorrtllm (based on their documentation, not from testing it out myself yet) is that they have a base image that we can utilize as long as we provide the prebuilt model weights - ref: https://github.com/triton-inference-server/tensorrtllm_backend/blob/main/README.md#run-the-pre-built-docker-container

This -gs.yaml is for the pathway for customers to pull in models that are stored in external storage (thus not built into the image), and it should rely on the base triton pathway, not the custom one. I'm assuming if we use the base image these packages are already pre-installed.

benchmarks/inference-server/triton/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+
+    # Use the environment variable to log in to Hugging Face CLI
+    huggingface-cli login --token $HUGGINGFACE_TOKEN
+
+
+
+
+    # Launch the servers (modify this depending on the number of GPU used and the exact path to the model repo)
+    mpirun --allow-run-as-root -n 1 /opt/tritonserver/bin/tritonserver \
```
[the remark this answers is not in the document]
kaushikmitr: in this particular example I am running two instances of triton server with different shared memory region prefix name as the engine was built for 2 GPUs. Will this still work without mpirun?

benchmarks/inference-server/triton/README.md
kaushikmitr: yes I thought about that but missed adding that. Makes sense.  let me define this in the .tfvars file.

benchmarks/inference-server/triton/manifest-templates/triton-tensorrtllm-inference-gs.tftpl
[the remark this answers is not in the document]
kaushikmitr: In this implementation, we utilize the base image nvcr.io/nvidia/tritonserver:23.10-trtllm-python-py3. However, we've made this a variable to allow users the flexibility to upgrade to newer versions as they become available. It may be beneficial to elaborate on this option further in the Readme for clarity.

Contrary to 'Option 1', outlined in the [Triton Inference Server documentation for TensorRT LL Model backend](https://github.com/triton-inference-server/tensorrtllm_backend/blob/main/README.md#option-1-launch-triton-server-within-triton-ngc-container), Option2 employs a base triton server image provided by Nvidia and specifies all configurations directly in the YAML file, eliminating the need to create a new image. We can intoduce  'Option 3' (or update 'Option 2'), which mirrors 'Option 1' but differs in that the model is loaded via Google Storage.

As of version 23.10, the SentencePiece and protobuf appears to be missing, and a Hugging Face login for llama is required. This is because the tokenizer is downloaded at runtime.

benchmarks/inference-server/triton/README.md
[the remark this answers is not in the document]
achandrasekar: Yes, "-n 1" essentially means we are running 1 instance. Multi-GPU should still work since a single Triton process can use multiple GPUs with Tensor parallelism.

benchmarks/inference-server/triton/manifest-templates/triton-tensorrtllm-inference-gs.tftpl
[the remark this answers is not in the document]
achandrasekar: Makes sense. If the plan is to make this yaml simpler, then defaulting the image_path to the Nvidia image by default and specifying the custom image as an additional option instead of providing them as two different options that the user has to choose between would be better.

benchmarks/inference-server/triton/manifest-templates/triton-tensorrtllm-inference-gs.tftpl
kaushikmitr: I will look into this in a subsequent PR.

benchmarks/inference-server/triton/manifest-templates/triton-tensorrtllm-inference-gs.tftpl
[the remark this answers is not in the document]
kaushikmitr: Yes, I made the gcs method as default and custoem image as an additional option

benchmarks/inference-server/triton/README.md
```diff
@@ -0,0 +1,198 @@
...
+    path = "~/.kube/${CLUSTER_NAME}-kube.config"
+  }
+}
+```
+
+#### [optional] set up secret token in Secret Manager
+
+A model may require a security token to access it. For example, Llama2 from HuggingFace is a gated model that requires a [user access token](https://huggingface.co/docs/hub/en/security-tokens). If the model you want to run does not require this, skip this step.
+
+If you followed steps from `../../infra/stage-2`, Secret Manager and the user access token should already be set up. Alternatively you can create a Kubernetes Secret to store your Hugging Face CLI token. You can do this from the command line with kubectl:
```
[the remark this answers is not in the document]
kaushikmitr: added the exttra comment. Will follow up with  added support for the secure way via secret manager

benchmarks/inference-server/triton/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+
+    # Use the environment variable to log in to Hugging Face CLI
+    huggingface-cli login --token $HUGGINGFACE_TOKEN
+
+
+
+
+    # Launch the servers (modify this depending on the number of GPU used and the exact path to the model repo)
+    /opt/tritonserver/bin/tritonserver \
```
achandrasekar, line 56: I'm wondering why we need to launch two different triton processes here. For multiple GPUs, just setting tensor_parallelism in the config should be enough right?

kaushikmitr: actually, I rested without mpirun and it seems to set tensor_parallelism we need to run via mpirun. Otherwise this assertion fails: 
Internal: unexpected error when creating modelInstanceState: [TensorRT-LLM][ERRO |
|                |         | R] Assertion failed: mpiSize == tp * pp (/app/tensorrt_llm/cpp/tensorrt_llm/runtime/worldConf |
|                |         | ig.cpp:80) 
https://github.com/NVIDIA/TensorRT-LLM/blob/728cc0044bb76d1fafbcaa720c403e8de4f81906/cpp/tensorrt_llm/runtime/worldConfig.cpp#L78C1-L78C63
in the example above tp = 2 and pp = 1 and mpiSize should be 2* 1 = 2

achandrasekar (approved):

benchmarks/inference-server/triton/README.md
annapendleton: Interesting - my guess is that the triton configuration wasn't updated as expected.

Triton has a couple files for config it looks like: 
1. https://nvidia.github.io/TensorRT-LLM/_cpp_gen/runtime.html#gptjsonconfig-h
2. https://nvidia.github.io/TensorRT-LLM/_cpp_gen/runtime.html#worldconfig-h

This is throwing an error at the world config not having the expected number of gpu's for the way the model was built.

I think we need to pass through the gpu's somehow - eg. --gpus = all to indicate use all the existing GPUs for the model maybe?
https://github.com/triton-inference-server/tensorrtllm_backend?tab=readme-ov-file#option-1-launch-triton-server-within-triton-ngc-container

Or I also see this option here to pass through the world-size value which correlates to the GPUs to use: 
https://github.com/triton-inference-server/tensorrtllm_backend?tab=readme-ov-file#option-3-launch-triton-server-within-the-triton-container-built-via-docker

You'll see the model weights are configured to use tp=4 in the conversion in the related example: https://github.com/triton-inference-server/tensorrtllm_backend?tab=readme-ov-file#prepare-tensorrt-llm-engines

annapendleton: Oh fun fact - that launch triton server script is backed by mpirun as well!
https://github.com/triton-inference-server/tensorrtllm_backend/blob/main/scripts/launch_triton_server.py#L67

I think Kaushik's observation is right that tensorrt_llm backend relies on mpirun for tensor parallelism.

annapendleton (approved):

PR #277, opened 2024-03-04
TPU Provisioner: Fix skipped scale ups & garbage collect zombie node pools

nstogner:
* Avoid scale ups when node pool is in a deletion state.
* Avoid scale ups when the Pod being reconciled is in a deletion state.
* Use owner UID for node pool names to avoid scale up failure when the scheduler has placed another workload on the same existing node pool.
* Run a loop that deletes node pools that have no Nodes, are in an errored state, and where the Pod that created the node pool no longer exists (the deletion reconciler would not see these b/c there are no Node objects).
* Update events to be more descriptive for easier debugging.
* Make Secure Boot configurable (defaults to same behavior).
* Make deletion delay configurable to ensure `make test` passes consistently.


tpu-provisioner/internal/controller/deletion_controller.go
```diff
@@ -142,17 +142,14 @@ func (r *DeletionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 	// and can be deleted.
 	lg.Info(fmt.Sprintf("Node pool %q passed deletion check twice. Ensuring Node Pool is deleted", nodePoolName))
-	r.Recorder.Event(&node, corev1.EventTypeNormal, EventDeletingNodePool, DeletingNodePoolEventMessage)
-	if err := r.Provider.DeleteNodePoolForNode(&node); err != nil {
+	if err := r.Provider.DeleteNodePoolForNode(&node, "no user Pods are running on any of the Nodes in this node pool"); err != nil {
 		if errors.Is(err, cloud.ErrDuplicateRequest) {
 			lg.Info("Ignoring duplicate request to delete node pool")
 			return ctrl.Result{}, nil
 		} else {
-			r.Recorder.Event(&node, corev1.EventTypeWarning, EventFailedDeletingNodePool, "Failed to delete Node Pool: "+err.Error())
```
danielvegamyhre, line 151: I see we are removing several multiple events we were previously emitting in different places, can you clarify why this is? Is the intent to consolidate the events to only the most useful, descriptive ones?

Also, why are we no longer ignoring the "Not found" error here?

tpu-provisioner/internal/controller/nodepool_garbage_collector.go
```diff
@@ -0,0 +1,94 @@
...
+	"time"
+
+	"github.com/GoogleCloudPlatform/ai-on-gke/tpu-provisioner/internal/cloud"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type NodePoolGarbageCollector struct {
```
danielvegamyhre, line 15: Can you add a comment with the info you included in the PR description about what this does?

e.g. `Deletes node pools that have no Nodes, are in an errored state, and where the Pod that created the node pool no longer exists (the deletion reconciler would not see these b/c there are no Node objects).`

danielvegamyhre (commented):
LGTM overall, left a few minor comments and questions

tpu-provisioner/internal/controller/deletion_controller.go
nstogner: The not found check lives in the cloud package now. I dont think this not-found check was working b/c it checks for k8s errors not GCP errors.

nstogner:
Updated to address comments. Also made deletion delay configurable to make sure `make test` passes consistently.

tpu-provisioner/internal/controller/deletion_controller.go
nstogner: PS: On events: yes, I refactored them to provide more meaningful information. To do this they needed to be fired from the cloud pkg.

tpu-provisioner/internal/controller/nodepool_garbage_collector.go
```diff
@@ -0,0 +1,98 @@
...
+
+			if np.CreatedForPod.Name == "" || np.CreatedForPod.Namespace == "" {
+				log.Info("skipping garbage collection of node pool, no pod reference")
+				continue
+			}
+
+			// Check if the Pod that triggered the Node Pool creation still exists.
+			err := g.Get(ctx, np.CreatedForPod, &v1.Pod{})
+			if err == nil {
+				log.Info("skipping garbage collection of node pool, pod still exists",
```
danielvegamyhre, line 65: What about the case where the pod which triggered node pool creation is still pending, waiting for the nodes to be ready but since the provisioning failed the nodes either don't exist or are in an ERROR state?

I encountered a case today where the pod was still pending on a NP which failed to provision, and the node pool was never deleted so the pod stayed pending and workload never made any progress.

danielvegamyhre: Since we are introducing this polling-based approach to NP deletion/clean up, I think we should consolidate the deletion logic to be completely encapsulated in this garbage collector, rather than split between a polling-based garbage collector and reconciliation-based deletion controller. This would simplify things and make it easier to reason about, and since we control the polling interval we won't be spamming the GKE API.

nstogner: > What about the case where the pod which triggered node pool creation is still pending, waiting for the nodes to be ready but the provisioning failed due to resource exhaustion so there are 0 nodes

The garbage collector would not delete the node pool in this case, and I would expect it goes into a repair state.

nstogner: > Since we are introducing this polling-based approach to NP deletion/clean up, I think we should consolidate the deletion logic to be completely encapsulated in this garbage collector, rather than split between a polling-based garbage collector and reconciliation-based deletion controller.

I think there might be some merit to this - but would want to explore in a later PR

nstogner:
Rebased @danielvegamyhre

danielvegamyhre (approved):

PR #284, opened 2024-03-05
Update TF dependencies to fix TF plan timing issue

umeshkumhar:
- Update frontend explicit depends_on which are not required
- Remove terraform provider that are not required in frontend
- update create_namespace to always true, because helm_release will take care if ns already exists and it wont attempt already exists error.

### DIFF
--- a/applications/rag/main.tf
+++ b/applications/rag/main.tf
@@ -88,17 +88,13 @@ provider "helm" {
   }
 }
 
-data "kubernetes_all_namespaces" "allns" {
-  provider   = kubernetes.rag
-  depends_on = [module.infra, data.google_container_cluster.default]
-}
 
 module "kuberay-operator" {
   source                 = "../../modules/kuberay-operator"
   providers              = { helm = helm.rag, kubernetes = kubernetes.rag }
   name                   = "kuberay-operator"
   project_id             = var.project_id
-  create_namespace       = !contains(data.kubernetes_all_namespaces.allns.namespaces, var.kubernetes_namespace)
+  create_namespace       = true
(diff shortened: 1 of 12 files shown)


applications/rag/main.tf
```diff
@@ -218,6 +214,5 @@ module "frontend" {
   url_domain_addr          = var.frontend_url_domain_addr
   url_domain_name          = var.frontend_url_domain_name
   members_allowlist        = var.frontend_members_allowlist
-
-  depends_on = [module.cloudsql, module.gcs, module.inference-server]
+  depends_on = [ module.gcs ]
```
hsachdevah, line 217: We'll still need depend on module.cloudsql
rag-frontend deployment has cloud-sql-proxy which requires the CONNECTION_NAME.

hsachdevah (changes requested):

applications/rag/main.tf
hsachdevah: It should be fine then.

hsachdevah (approved):

artemvmin:
Could you please document what dependencies are being updated?

Additionally could you please document all the other bug fixes and changes being made? Generally, a person reading through the commit log and/or PR list shouldn't have to read the code to understand the changes that have been made. This also helps the reviewer understand what your intention is with the PR. For example, the following don't seem to be related to your PR title `Update TF dependencies to fix TF plan timing issue`:

- `ClientIP` -> `ClusterIP` typo
- Namespace is now always being created in some modules
  - Do we need this and does this have to be done in this PR? Let's reduce the scope to make sure we fix the timing issue completely.

PR #295, opened 2024-03-06
Cleanup Cloud SQL DB creds for notebook and frontend

imreddy13:
1) Use a k8s secret for DB creds
2) Mount k8s secret in volumes for Ray & frontend pods
3) Read the mounted secret volume from the Ray job script in the notebook and from the frontend container
4) Remove unecessesary DB cred variables from the notebook
5) Fix bug with NLP/DLP imports


applications/rag/README.md
```diff
@@ -135,18 +135,20 @@ This step generates the vector embeddings for your input dataset. Currently, the
...
 
-8. Replace the kaggle username and api token in 2nd cell with your credentials (can be found in the `kaggle.json` file created by Step 6):
-    * `os.environ['KAGGLE_USERNAME']`
-    * `os.environ['KAGGLE_KEY']`
+6. Replace the variables in the 1st cell with custom values:
+    * Kaggle credentials (can be found in the `kaggle.json` file created by Step 5):
+        * `os.environ['KAGGLE_USERNAME']`
+        * `os.environ['KAGGLE_KEY']`
+    * DB credentials:
+        * `DB_REGION` = ${CLUSTER_REGION:?} # Note this needs to be the GCE region the instance is in (not zone).
```
artemvmin, line 146: This implies it's being pasted into shell. Let's just enumerate the variables that need to be filled in:

* DB credentials:
  * `DB_REGION`
  * `PROJECT_ID`
  * `DB_PASS` (from step 1)

Additionally:
* Drop `DB_USER`, as most users won't change it and it has a good default.
* "GCE region (not zone) of the instance" is already spelled out in the notebook. Since they will be reading that comment anyways, I wouldn't duplicate the note in both places.

artemvmin (approved):

applications/rag/README.md
```diff
@@ -135,18 +135,20 @@ This step generates the vector embeddings for your input dataset. Currently, the
...
+        * `PROJECT_ID`= <project_id>
+        * `DB_USER`: `rag-user-notebook`
+        * `DB_PASS`: password from step 1
```
bjornsen, line 149: Can we convert the secrets in this document into Kubernetes Secrets and mount them into the various jobs that need them? This is safer than direct setting and envvars. It'll mount them as files in your pod.

https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#create-a-pod-that-has-access-to-the-secret-data-through-a-volume

applications/rag/main.tf
```diff
@@ -169,6 +170,8 @@ module "kuberay-cluster" {
   enable_tpu             = local.enable_tpu
   autopilot_cluster      = local.enable_autopilot
   google_service_account = var.ray_service_account
+  db_secret_name         = module.cloudsql.db_secret_name
```
blackzlq, line 173: Also update the module in ray

applications/rag/frontend/container/main.py
```diff
@@ -41,8 +41,14 @@
 # initialize parameters
 INFERENCE_ENDPOINT=os.environ.get('INFERENCE_ENDPOINT', '127.0.0.1:8081')
 INSTANCE_CONNECTION_NAME = os.environ.get('INSTANCE_CONNECTION_NAME', '')
-DB_USER = os.environ.get('DB_USER', '')
-DB_PASS = os.environ.get('DB_PASSWORD', '')
+
+db_username_file = open("/etc/secret-volume/username", "r")
```
blackzlq, line 45: Have you tested note book? do we need to add any other permission?

imreddy13: tested E2E and it works with this!

applications/rag/main.tf
imreddy13: Good catch!

Instead, I made all the secrets optional. TIL: https://stackoverflow.com/questions/48222827/optional-volume-secret-volume-in-kubernetes

Tested that /applications/ray deploys successfully and the Ray pods come up.

applications/rag/frontend/container/rai/dlp_filter.py
```diff
@@ -14,7 +14,7 @@
 
 import os
 import google.cloud.dlp
-from .retry import retry
+from . import retry
```
bjornsen, line 17: Why change back to relative imports?

applications/rag/frontend/container/rai/nlp_filter.py
```diff
@@ -14,7 +14,7 @@
 
 import os
 import google.cloud.language as language
-from .retry import retry
+from . import retry
```
bjornsen, line 17: Same question about relative imports

modules/kuberay-cluster/kuberay-autopilot-values.yaml
bjornsen, line 1: Four different kuberay-cluster files strikes again 😢

applications/rag/frontend/container/rai/nlp_filter.py
imreddy13: answered above

modules/kuberay-cluster/kuberay-autopilot-values.yaml
imreddy13: Sigh :(

applications/rag/frontend/container/rai/dlp_filter.py
blackzlq: retry is under the same level of the folder

bjornsen (approved):

blackzlq (approved):

PR #296, opened 2024-03-06
Add github actions for tf lint check

umeshkumhar:
- Add github actions for TF fmt, TF validate and TF init

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,7 +18,6 @@ jobs:
       - name: Terraform fmt
         id: fmt
         run: terraform fmt -check -recursive
-        continue-on-error: false
 
       - name: Terraform Init
         id: init
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ on:
     branches:
       - main
 jobs:
-  Terraform-Fmt-Check:
+  Terraform-Lint-Check:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,36 @@
+name: Terraform CI
+on: 
+  push:
+    branches:
+      - main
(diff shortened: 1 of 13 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: "1.5.7"
+
+      - name: Terraform fmt
+        id: fmt
+        run: terraform fmt -check -recursive
+        continue-on-error: false
```
blackzlq, line 21: why we have continue-on-error only on fmt?

umeshkumhar: its not required though,, its default behavior to fail on error

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,35 @@
...
+        run: terraform fmt -check -recursive
+
+      - name: Terraform Init
+        id: init
+        run: |
+          terraform -chdir=applications/rag init 
+          terraform -chdir=applications/ray init 
+          terraform -chdir=applications/jupyter init 
+
+      - name: Terraform Validate
```
andrewsykim, line 29: We already do this check in cloud build

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  Terraform-Lint-Check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: "1.5.7"
+
+      - name: Terraform fmt
```
andrewsykim, line 18: Can you just add this lint check to cloudbuild.yaml and remove the github actions workflow? I don't think we're at a point where we want to manage both cloudbuild and github actions for presubmit.

andrewsykim: (also because cloudbuild is a blocking pre-submit now)

PR #298, opened 2024-03-06
locust custom metrics tgi support

annapendleton:


benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py
```diff
@@ -23,12 +23,12 @@ def calculate_average_tokens(self):
         avg_sent = sum(self.tokens_sent) / len(self.tokens_sent) if self.tokens_sent else 0
         avg_received = sum(self.tokens_received) / len(self.tokens_received) if self.tokens_received else 0
         avg_test_time = sum(self.test_time) / len(self.test_time) if self.tokens_sent else 0
-        
-        return avg_sent, avg_received, avg_test_time
+        avg_output_token_latency = avg_received / avg_test_time
```
achandrasekar, line 26: This might not be the accurate way to calculate output token latency. Basically we need to calculate latency per request and output token latency per request and then average all of the output token latencies from all the requests instead of calculating them cumulatively because latency is a request level metric.

annapendleton: A good example of why I shouldn't code past midnight 🤦 my brain somehow convinced me the math was the same as the for loop method I originally had. updated with for loop, lmk if that reflects the intended behavior

PR #299, opened 2024-03-06
Upgrade Ray version for Autopilot; shrink worker resource allocation

artemvmin:
- Ray version 2.7.1 appears to fix some Ray Autoscaler bugs
- Lower resource allocation to better utilize node capacity (recommended 1:1 matching between k8s and Ray nodes):
  - Head: `c3-standard-4`
  - CPU worker:`c3-standard-8`
  - GPU worker:`g2-standard-8`
- Pretty print Ray job info in Jupyter notebook
- Fix README wording

### DIFF
--- a/modules/kuberay-cluster/kuberay-autopilot-values.yaml
+++ b/modules/kuberay-cluster/kuberay-autopilot-values.yaml
@@ -1,4 +1,4 @@
-# Copyright 2023 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -22,7 +22,7 @@
 image:
   # Replace this with your own image if needed.
   repository: rayproject/ray
-  tag: 2.6.1-py310-gpu
+  tag: 2.9.3-py310-gpu
   pullPolicy: IfNotPresent
 
 nameOverride: "kuberay"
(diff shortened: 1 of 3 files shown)


modules/kuberay-cluster/kuberay-autopilot-values.yaml
```diff
@@ -22,7 +22,7 @@
 image:
   # Replace this with your own image if needed.
   repository: rayproject/ray
-  tag: 2.6.1-py310-gpu
+  tag: 2.9.3-py310-gpu
```
imreddy13, line 25: should we bump it for Standard as well to keep things consistent?

artemvmin: That's going to be a larger, possibly breaking change, that affects multiple applications. I would rather do that in a separate PR.

imreddy13 (approved):

modules/kuberay-cluster/kuberay-autopilot-values.yaml
saikat-royc: this needs a corresponding change to the notebook where we pip install ray ==2.7.x . We need to align the ray versions ?

Has this change been validate with the notebook e2e?

imreddy13: +1 to what Saikat said, my bad for missing this in the review. the notebook uses the same ray version as the standard kuberay config, does this work for AP?

PR #301, opened 2024-03-06
Add cloudsql dependency to rag frontend

artemvmin:
Without this, the frontend pod cloud-sql-proxy container crash loops with auth errors for the entire 10 minutes that the CloudSQL instance is starting up.


umeshkumhar:
This is fixed here #302

imreddy13 (changes requested):
revert this?

artemvmin:
Thanks for the fix.

PR #303, opened 2024-03-06
Optimize CI

hsachdevah:
- Upgrade machineType for Cloudbuild Instance
- Pin workload ID TF module version

### DIFF
--- a/modules/kuberay-operator/kuberay.tf
+++ b/modules/kuberay-operator/kuberay.tf
@@ -32,8 +32,6 @@ module "kuberay-workload-identity" {
   project_id          = var.project_id
   roles               = ["roles/cloudsql.client", "roles/monitoring.viewer"]
 
-  automount_service_account_token = true
-
   depends_on = [helm_release.kuberay-operator]
 }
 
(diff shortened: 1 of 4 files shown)


modules/kuberay-operator/kuberay.tf
```diff
@@ -25,6 +25,7 @@ resource "helm_release" "kuberay-operator" {
 
 module "kuberay-workload-identity" {
   source              = "terraform-google-modules/kubernetes-engine/google//modules/workload-identity"
+  version             = "30.0.0"
```
imreddy13, line 28: can you add a comment mentioning why we pin this?
can we also pin this in frontend and jupyter?

search for all refs: https://github.com/search?q=repo%3AGoogleCloudPlatform%2Fai-on-gke%20terraform-google-modules%2Fkubernetes-engine%2Fgoogle%2F%2Fmodules%2Fworkload-identity&type=code

hsachdevah: Fixed

umeshkumhar (approved):

modules/kuberay-operator/kuberay.tf
```diff
@@ -32,8 +32,6 @@ module "kuberay-workload-identity" {
   project_id          = var.project_id
   roles               = ["roles/cloudsql.client", "roles/monitoring.viewer"]
 
-  automount_service_account_token = true
```
artemvmin, line 34: FYI this broke the RAG application. See: https://github.com/GoogleCloudPlatform/ai-on-gke/commit/6b844bcfa3ead58a10ded54c215e61e856c8ffae#r139575280

PR #314, opened 2024-03-07
Refactor IAP

blackzlq:
Refactor IAP module

Tested-by: zlq on AP cluster and standard

### DIFF
--- a/applications/jupyter/main.tf
+++ b/applications/jupyter/main.tf
@@ -77,6 +77,16 @@ module "namespace" {
   create_namespace = true
 }
 
+# IAP Section: Enabled the IAP service
+resource "google_project_service" "project_service" {
+  count   = var.add_auth ? 1 : 0
+  project = var.project_id
+  service = "iap.googleapis.com"
+
+  disable_dependent_services = false
+  disable_on_destroy         = false
+}
+
 # Creates jupyterhub
 module "jupyterhub" {
   source                            = "../../modules/jupyter"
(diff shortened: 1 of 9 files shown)


applications/jupyter/main.tf
```diff
@@ -77,6 +77,16 @@ module "namespace" {
   create_namespace = true
 }
 
+# IAP Section: Enabled the IAP service
+resource "google_project_service" "project_service" {
```
umeshkumhar, line 81: can this be moved into iap module

blackzlq: Discussed offline, we will do another test later. From my option this is project level, should be enabled once. and Umesh thinks it is good for code maintain and generality of IAP module, also the terraform will auto correct it.

modules/iap/iap.tf
```diff
@@ -12,152 +12,98 @@
...
   provider     = google-beta
   project      = var.project_id
-  name         = "frontend-address-${random_string.random.result}"
+  name         = "${var.app_name}-address-${random_string.random.result}"
   address_type = "EXTERNAL"
   ip_version   = "IPV4"
 }
 
 # Helm Chart IAP
 resource "helm_release" "iap_frontend" {
```
umeshkumhar, line 53: rename iap_frontend to iap

umeshkumhar (approved):

modules/iap/iap.tf
blackzlq: Good catch!

PR #320, opened 2024-03-08
Decouple the TGI PodMonitoring from HPA

laoj2:
Today, we only deploy a PodMonitoring object if HPA is enabled to autoscale based on TGI metrics. But, TGI metrics are useful for Observability anyway, so let's decouple this from the HPA deployment.


benchmarks/inference-server/text-generation-inference/main.tf
```diff
@@ -84,7 +84,6 @@ resource "kubernetes_manifest" "hpa-cpu" {
 }
 
 resource "kubernetes_manifest" "tgi-pod-monitoring" {
-  count = local.custom_metrics_enabled ? 1 : 0
```
annapendleton, line 87: Should we decouple the yaml files from the same directory as well? I guess I'm thinking maybe it makes sense to move the podmonitoring yaml to either a separate monitoring folder or the existing tgi manifests folder to indicate its always deployed.

annapendleton (changes requested):

benchmarks/inference-server/text-generation-inference/main.tf
laoj2: Yes, that's a good idea. 

I kept the DCGM PodMonitoring in hpa-templates because that one is specific for HPA (due to the metric renaming hack)

annapendleton (approved):

PR #328, opened 2024-03-11
Improvements to Cloud SQL & cluster creation

imreddy13:
1) Pass already constructed cloud SQL instance connection name to Ray workers so users don't have to set it in the notebook
2) Set Cloud SQL region in frontend
3) Fix bug with cluster creation where random zones without GPU availability were being selected
4) Reduce SA name length (gets truncated when IM deployment prefix is added for Marketplace)

### DIFF
--- a/applications/rag/main.tf
+++ b/applications/rag/main.tf
@@ -225,6 +225,7 @@ module "frontend" {
   namespace                     = var.kubernetes_namespace
   inference_service_endpoint    = module.inference-server.inference_service_endpoint
   cloudsql_instance             = module.cloudsql.instance
+  cloudsql_instance_region      = var.cloudsql_instance_region
   db_secret_name                = module.cloudsql.db_secret_name
   dataset_embeddings_table_name = var.dataset_embeddings_table_name
 
--- a/applications/rag/workloads.tfvars
+++ b/applications/rag/workloads.tfvars
@@ -30,16 +30,16 @@ cloudsql_instance_region = "us-central1"
 # Creates a google service account & k8s service account & configures workload identity with appropriate permissions.
 # Set to false & update the variable `ray_service_account` to use an existing IAM service account.
 create_ray_service_account      = true
-ray_service_account             = "ray-system-account"
+ray_service_account             = "ray-sa"
 enable_grafana_on_ray_dashboard = false
 # Creates a google service account & k8s service account & configures workload identity with appropriate permissions.
 # Set to false & update the variable `rag_service_account` to use an existing IAM service account.
 create_rag_service_account = true
-rag_service_account        = "rag-system-account"
+rag_service_account        = "rag-sa"
 
 # Creates a google service account & k8s service account & configures workload identity with appropriate permissions.
 # Set to false & update the variable `jupyter_service_account` to use an existing IAM service account.
-jupyter_service_account = "jupyter-system-account"
+jupyter_service_account = "jupyter"
 
 ## Embeddings table name - change this to the TABLE_NAME used in the notebook.
 dataset_embeddings_table_name = "googlemaps_reviews_db"
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -62,7 +62,9 @@ locals {
   subnetwork_name = var.create_network ? module.custom-network[0].subnets_names[0] : var.subnetwork_name
   region          = length(split("-", var.cluster_location)) == 2 ? var.cluster_location : ""
   regional        = local.region != "" ? true : false
-  zone            = length(split("-", var.cluster_location)) > 2 ? split(",", var.cluster_location) : []
+  # zone needs to be set even for regional clusters, otherwise this module picks random zones that don't have GPU availability:
+  # https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/af354afdf13b336014cefbfe8f848e52c17d4415/main.tf#L46 
+  zone = length(split("-", var.cluster_location)) > 2 ? split(",", var.cluster_location) : split(",", local.gpu_l4_t4_location[local.region])
   # Update gpu_pools with node_locations according to region and zone gpu availibility, if not provided
   gpu_pools = [for elm in var.gpu_pools : (local.regional && contains(keys(local.gpu_l4_t4_location), local.region) && elm["node_locations"] == "") ? merge(elm, { "node_locations" : local.gpu_l4_t4_location[local.region] }) : elm]
 }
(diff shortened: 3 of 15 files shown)


modules/kuberay-cluster/kuberay-autopilot-values.yaml
```diff
@@ -73,10 +73,8 @@ head:
       value: http://grafana:80
     - name: RAY_PROMETHEUS_HOST
       value: http://frontend:9090
-    - name: PROJECT_ID
-      value: ${project_id}
```
blackzlq, line 77: Why remove project id?

applications/rag/main.tf
```diff
@@ -182,7 +182,7 @@ module "kuberay-cluster" {
   providers              = { helm = helm.rag, kubernetes = kubernetes.rag }
   project_id             = var.project_id
   namespace              = var.kubernetes_namespace
-  enable_gpu             = true
+  enable_gpu             = false
```
blackzlq, line 185: why we disabled gpu?

infrastructure/main.tf
```diff
@@ -62,7 +62,9 @@ locals {
   subnetwork_name = var.create_network ? module.custom-network[0].subnets_names[0] : var.subnetwork_name
   region          = length(split("-", var.cluster_location)) == 2 ? var.cluster_location : ""
   regional        = local.region != "" ? true : false
-  zone            = length(split("-", var.cluster_location)) > 2 ? split(",", var.cluster_location) : []
+  # zone needs to be set even for regional clusters, otherwise this module picks random zones that don't have GPU availability:
+  # https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/blob/af354afdf13b336014cefbfe8f848e52c17d4415/main.tf#L46 
+  zone            = length(split("-", var.cluster_location)) > 2 ? split(",", var.cluster_location) : split(",", local.gpu_l4_t4_location[local.region])
```
andrewsykim, line 67: Wow, good catch. It's kind of unexpected that it does this

andrewsykim (approved):

imreddy13:
bypassing pr-review-trigger after updating 2 files with tf fmt

passed on previous iteration: https://github.com/GoogleCloudPlatform/ai-on-gke/pull/328/checks?check_run_id=22591229809

applications/rag/workloads.tfvars
```diff
@@ -30,16 +30,16 @@ cloudsql_instance_region = "us-central1"
...
 # Creates a google service account & k8s service account & configures workload identity with appropriate permissions.
 # Set to false & update the variable `rag_service_account` to use an existing IAM service account.
 create_rag_service_account = true
-rag_service_account        = "rag-system-account"
+rag_service_account        = "rag-sa"
 
 # Creates a google service account & k8s service account & configures workload identity with appropriate permissions.
 # Set to false & update the variable `jupyter_service_account` to use an existing IAM service account.
-jupyter_service_account = "jupyter-system-account"
+jupyter_service_account = "jupyter"
```
blackzlq, line 42: Why only jupyter has no sa?

infrastructure/main.tf
```diff
@@ -62,7 +62,9 @@ locals {
   subnetwork_name = var.create_network ? module.custom-network[0].subnets_names[0] : var.subnetwork_name
   region          = length(split("-", var.cluster_location)) == 2 ? var.cluster_location : ""
   regional        = local.region != "" ? true : false
```
blackzlq, line 64: In this case regional should be false?

modules/kuberay-cluster/main.tf
```diff
@@ -31,37 +32,33 @@ resource "helm_release" "ray-cluster" {
...
-      security_context    = local.security_context
-      secret_name         = var.db_secret_name
-      project_id          = var.project_id
-      db_region           = var.db_region
+      gcs_bucket                        = var.gcs_bucket
+      k8s_service_account               = var.google_service_account
+      grafana_host                      = var.grafana_host
+      security_context                  = local.security_context
+      secret_name                       = var.db_secret_name
+      cloudsql_instance_connection_name = local.cloudsql_instance_connection_name
```
blackzlq, line 40: And we still manually type the instance in jupyter hub. will this cause any mis-configuration?

modules/kuberay-cluster/variables.tf
```diff
@@ -29,6 +29,12 @@ variable "db_region" {
   default     = "us-central1"
 }
 
+variable "cloudsql_instance_name" {
```
blackzlq, line 32: I guess we should also make here to be the var and pass it from application layer: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/modules/cloudsql/main.tf#L64C16-L64C35

modules/kuberay-cluster/main.tf
imreddy13: no, we don't need to - see the associated notebook change

modules/kuberay-cluster/variables.tf
imreddy13: already exists: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/modules/cloudsql/variables.tf#L26

modules/kuberay-cluster/kuberay-values.yaml
```diff
@@ -173,10 +171,8 @@ worker:
   # containerEnv specifies environment variables for the Ray container,
   # Follows standard K8s container env schema.
   containerEnv:
-    - name: PROJECT_ID
-      value: ${project_id}
```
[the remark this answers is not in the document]
imreddy13: it's not. i explicitly added it for this purpose: https://github.com/GoogleCloudPlatform/ai-on-gke/pull/295

infrastructure/main.tf
imreddy13: this is so complicated.
I think region = var.cluster_location if it's a regional cluster.
so local.region != "" is true for regional clusters. so this is correct.

@umeshkumhar this naming is super confusing

applications/rag/workloads.tfvars
imreddy13: SA name has 6 character minimum, 30 character maximum.
we prefix with IM deployment name for marketplace, prefer to keep it small so it doesn't get truncated.

infrastructure/main.tf
umeshkumhar: Just to clarify.. we take **cluster_location** input only.

Then we check if user provided region or zone by splitting on hypen. Once done, we check if region is set that means cluster will be regional, so make it true, else false.

PR #331, opened 2024-03-11
add max-concurrent-requests as option in TGI server

annapendleton:
Validated via tf deployment:

kubectl logs -n benchmark tgi-66f66b7b6f-np2nw

output shows max_concurrent_requests: 1000 in tgi server config

### DIFF
--- a/benchmarks/inference-server/text-generation-inference/variables.tf
+++ b/benchmarks/inference-server/text-generation-inference/variables.tf
@@ -58,6 +58,18 @@ variable "gpu_count" {
   }
 }
 
+variable "max_concurrent_requests" {
+  description = "Max concurrent requests allowed for TGI to handle at once. TGI will drop all requests once it hits this max-concurrent-requests limit."
+  type        = number
+  nullable    = false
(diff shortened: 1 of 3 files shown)


laoj2 (approved):

benchmarks/inference-server/text-generation-inference/variables.tf
```diff
@@ -58,6 +58,17 @@ variable "gpu_count" {
   }
 }
 
+variable "max_concurrent_requests" {
+  description = "Max concurrent requests allowed for TGI to handle at once."
```
achandrasekar, line 62: Would be good to add to the description that TGI will drop all requests once it hits the max-concurrent-requests limit. Also, once we find a good default that works with the default benchmarking setup (1000 or whatever that number is), we should set that as default instead of 128.

annapendleton: great idea, updated the description!

+1, will update the default post-experimentation :) added todo for now

achandrasekar (approved):

PR #337, opened 2024-03-12
Add networkpolicy for kuberay clusters

bjornsen:
* Add default network policy for kuberay clusters
* Turn on Dataplanev2 for GKE Standard clusters
* Add an option to disable network policy


applications/rag/foo.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: networking.k8s.io/v1
```
blackzlq, line 1: What is this foo.yaml used for?

applications/rag/variables.tf
```diff
@@ -366,3 +366,9 @@ variable "goog_cm_deployment_name" {
   type    = string
   default = ""
 }
+
+variable "disable_ray_cluster_network_policy" {
```
blackzlq, line 370: If we update here, do we need to update application/ray?

modules/gke-standard-private-cluster/variables.tf
```diff
@@ -127,3 +127,9 @@ variable "gpu_pools" {
 variable "tpu_pools" {
   type = list(map(any))
 }
+
+variable "datapath_provider" {
+  description = "Default to Dataplanev2"
+  type        = string
+  default     = "ADVANCED_DATAPATH"
```
blackzlq, line 134: what is the other options? maybe also update the read me in rag or ray?

modules/kuberay-cluster/main.tf
```diff
@@ -73,3 +73,100 @@ data "kubernetes_service" "head-svc" {
...
+    pod_selector {
+      match_labels = {
+        "ray.io/is-ray-node" : "yes"
+      }
+    }
+
+    ingress {
+      # Ray Client server
+      ports {
+        port     = "10001"
```
blackzlq, line 95: Can we get the port from other resource? since we already import the ray cluster above

modules/kuberay-cluster/variables.tf
```diff
@@ -118,8 +118,33 @@ variable "security_context" {
   }
 }
 
+# By default, intra-namespace ingress is allowed to let the cluster talk to itself
+variable "network_policy_allow_namespaces_by_label" {
+  description = "Namespaces allowed to access this kuberay cluster"
+  type        = list(map(string))
+  default     = []
```
blackzlq, line 125: Do we have any example because list of map string is hard to understand

applications/rag/variables.tf
bjornsen: I didn't bother to update it in `application/ray` since we want that to have networkpolicy by default.

applications/rag/foo.yaml
bjornsen: Oops, how did that get in there?

modules/gke-standard-private-cluster/variables.tf
bjornsen: How are we autogenerating the READMEs for modules? I'm happy to regenerate that. What would you suggest to add into the higher level READMEs? This is just setting a default, lower level option defined in the GKE cluster modules:

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#datapath_provider

modules/kuberay-cluster/main.tf
bjornsen: Is there a way to pull these values from the helm_release via outputs? I don't think I have access to these ports unless I set them as variables on the module and push them through to the helm chart.

modules/gke-standard-private-cluster/variables.tf
imreddy13: manual generation

imreddy13 (approved):

applications/rag/variables.tf
bjornsen: I'd like to see us work through the implications of this in the Ray application. gmp-operator should be made more resilient as opposed to discarding security. We can always turn it off in the next commit if it's too much of a hassle.

PR #339, opened 2024-03-12
Remove enable_TPU as default

imreddy13:


applications/ray/variables.tf
```diff
@@ -85,6 +85,11 @@ variable "enable_gpu" {
   default = false
 }
 
+variable "enable_tpu" {
```
bjornsen, line 88: Add description.

bjornsen (approved):

applications/jupyter/main.tf
```diff
@@ -53,7 +53,6 @@ locals {
   private_cluster       = var.create_cluster ? var.private_cluster : data.google_container_cluster.default[0].private_cluster_config.0.enable_private_endpoint
   cluster_membership_id = var.cluster_membership_id == "" ? var.cluster_name : var.cluster_membership_id
   enable_autopilot      = var.create_cluster ? var.autopilot_cluster : data.google_container_cluster.default[0].enable_autopilot
-  enable_tpu            = var.create_cluster ? true : data.google_container_cluster.default[0].enable_tpu
```
umeshkumhar, line 56: can we set false here only..

so in case if cluster has tpu enabled.. it will be marked true automatically.

PR #341, opened 2024-03-12
Increase locust user task and connection timeout

annapendleton:
Increase locust user task and connection timeout to handle long-running requests.

Currently defaults are:
user_task -> 5 min before master kills the user task
http requests connection timeout -> 60 seconds before connection timeout error (explains the 0 return from locust)

Setting extremely high default for now (3 hrs)

Will turn into tfvars variable if possible (only unknown in that is that http request connection_timeout is a user class variable, and have to confirm that it will respect arguments piped through to tasks.py file before it's instantiated.)

### DIFF
--- a/benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
+++ b/benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
@@ -33,6 +33,7 @@ metric_collector = TokenMetricCollector()
 
 logging.basicConfig(level=logging.INFO)
 
+
 def load_test_prompts():
     """Loads test prompts from a local file location."""
     with open("locust-tasks/filtered_prompts.txt") as f:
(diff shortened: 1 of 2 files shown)


benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
```diff
@@ -99,39 +100,43 @@ def generate_request(prompt):
...
         raise ValueError(f"Unknown backend: {backend}")
     return number_of_input_tokens, number_of_output_tokens
 
 
-
 class BenchmarkUser(FastHttpUser):
     weight = 1
+    # Connection_timeout default is 60s. For inferencing workloads with a large payload
+    # this timeout can be too short. Increasing timeout to large amount.
+    # TODO: turn timeout into a variable.
```
laoj2, line 131: is this used anywhere? or is it just a placeholder for that TODO?

laoj2 (approved):

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
annapendleton: placeholder for a todo :) no tracking issue at this point

PR #342, opened 2024-03-13
Fix SQL name and region

umeshkumhar:
- Pass SQL instance name to module
- Use cluster_location as sql instance region, if not specified.


applications/rag/variables.tf
```diff
@@ -273,7 +273,7 @@ variable "cloudsql_instance" {
 variable "cloudsql_instance_region" {
   type        = string
   description = "GCP region for CloudSQL instance"
-  default     = "us-central1"
+  default     = ""
```
imreddy13, line 276: either delete this line and force the user to specify a region, or leave a meaningful default

imreddy13 (approved):

PR #348, opened 2024-03-13
Add Resource Quotas to Ray namespace

bjornsen:


blackzlq (approved):

imreddy13 (approved):

modules/kuberay-cluster/main.tf
```diff
@@ -167,3 +168,15 @@ resource "kubernetes_network_policy" "kuberay-cluster-allow-network-policy" {
     policy_types = ["Ingress"]
   }
 }
+
+# Assign resource quotas to Ray namespace to ensure that they don't overutilize resources
+resource "kubernetes_resource_quota" "ray_namespace_resource_quota" {
```
umeshkumhar, line 173: this resource block create quota on overall namespace.. and it will impact other dependent deployment and services.

Either add the match expression so this applies to only selected resources..

or take this out of this module..and add at namespace level.

bjornsen: There is no match expression for resource quotas. They always apply at the namespace level though you can scope them to priority classes (which we haven't defined). These defaults are left intentionally high so that they won't interfere with anything in our demo. If you're concerned, the right thing to do is to split Ray out into a separate namespace.

modules/kuberay-cluster/variables.tf
```diff
@@ -172,3 +172,15 @@ variable "disable_network_policy" {
   type        = bool
   default     = false
 }
+
+# These default resource quotas are set intentionally high as an example that won't be limiting for most clusters.
+# Consult https://kubernetes.io/docs/concepts/policy/resource-quotas/ for additional quotas that may be set.
+variable "resource_quotas" {
```
andrewsykim, line 178: Resource Quota is deprecated starting in GKE v1.28 right?

modules/kuberay-cluster/main.tf
umeshkumhar: okay agree.. lets move this out of kube-ray-cluster module.. and place it with rag/main.tf so it is configurable at that level.. rather then configuring it while calling kuberay-cluster module.

bjornsen: This isn't for RAG. This is for a blog post about Ray hardening that I'm writing. I'm testing against RAG because RAG is using Ray.

modules/kuberay-cluster/variables.tf
bjornsen: [citation needed]

PR #357, opened 2024-03-14
fix issue with big images for gke-disk-image-builder

ctrlok:
We tried to create a disk image for a big docker image (>50GB) and faced issues. 

The main problem is that output to the serial console is super verbose and too slow, so we disabled logs from the docker pull command. 

The second problem is that we need much more space for the host system to pull the image and unpack it. For that, we introduced an additional option.

### DIFF
--- a/gke-disk-image-builder/cli/main.go
+++ b/gke-disk-image-builder/cli/main.go
@@ -55,6 +55,7 @@ func main() {
 	serviceAccount := flag.String("service-account", "default", "Service Account email assigned to the GCE instance used for creating the disk image.")
 	diskType := flag.String("disk-type", "pd-ssd", "disk type to generate the disk image")
 	diskSizeGb := flag.Int64("disk-size-gb", 60, "disk size to unpack container images")
+	instanceDiskSizeGb := flag.Int64("instance-disk-size-gb", 100, "disk size of the GCE instance used for disk image creation")
 	gcpOAuth := flag.String("gcp-oauth", "", "path to GCP service account credential file")
 	imagePullAuth := flag.String("image-pull-auth", "None", "auth mechanism to pull the container image, valid values: [None, ServiceAccountToken].\nNone means that the images are publically available and no authentication is required to pull them.\nServiceAccountToken means the service account oauth token will be used to pull the images.\nFor more information refer to https://cloud.google.com/compute/docs/access/authenticate-workloads#applications")
 	timeout := flag.String("timeout", "20m", "Default timout for each step, defaults to 20m")
@@ -110,6 +111,7 @@ func main() {
 		ServiceAccount:        *serviceAccount,
 		DiskType:              *diskType,
 		DiskSizeGB:            *diskSizeGb,
+		InstanceDiskSizeGB:    *instanceDiskSizeGb,
 		GCPOAuth:              *gcpOAuth,
 		Network:               fmt.Sprintf("projects/%s/global/networks/%s", *projectName, *network),
 		Subnet:                fmt.Sprintf("projects/%s/regions/%s/subnetworks/%s", *projectName, regionForZone(*zone), *subnet),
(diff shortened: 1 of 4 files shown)


elfinhe:
Thanks Seva. Please help unblock by review and sign the CLA in https://github.com/GoogleCloudPlatform/ai-on-gke/pull/357/checks?check_run_id=22664449967

ctrlok:
@elfinhe, thanks; I initially had it signed from the wrong account. It should be better now.

gke-disk-image-builder/cli/main.go
```diff
@@ -55,6 +55,7 @@ func main() {
 	serviceAccount := flag.String("service-account", "default", "Service Account email assigned to the GCE instance used for creating the disk image.")
 	diskType := flag.String("disk-type", "pd-ssd", "disk type to generate the disk image")
 	diskSizeGb := flag.Int64("disk-size-gb", 60, "disk size to unpack container images")
+	instanceDiskSizeGb := flag.Int64("instance-disk-size-gb", 100, "disk size of the GCE instance used for disk image creation")
```
elfinhe, line 58: please add logic for if instanceDiskSizeGb is not specified, use the value from `diskSizeGb`

gke-disk-image-builder/script/startup.sh
```diff
@@ -72,9 +72,9 @@ function pull_images() {
   for param in "$@"; do
     echo Start pulling $param ...
     if [ "$OAUTH_MECHANISM" == "none" ]; then
-      sudo ctr -n k8s.io image pull $param
+      sudo ctr -n k8s.io image pull $param > /dev/null
```
elfinhe, line 75: maybe we can at least truncate the output and still print the first 300 lines?

elfinhe: The output here will be useful for debugging image pulling issues.

ctrlok:
@elfinhe hey

I pushed a new commit to address the issues you mentioned. 

# Why is a new type defined instead of a simple check 

I defined a new type instead of using an absurd value and `if` check mainly because that absurd value then appears in the `--help` output. 

```
  -instance-disk-size-gb int
    	disk size of the GCE instance used for disk image creation. (default -1) 
```

With this struct defined, we can have a nice message instead.

# Why (head; cat) 

`ctr` [didn't support](https://github.com/containerd/containerd/issues/9664) `pull` command without a progress bar, so I was forced to truncate everything after line 300. We can track the issue and rewrite it back once it is done.

ctrlok:
@elfinhe hey! Are there any updates? Thanks!

gke-disk-image-builder/cli/main.go
```diff
@@ -65,6 +66,9 @@ func main() {
 	flag.Var(&imageLabels, "image-labels", "labels tagged to the disk image. This flag can be specified multiple times. The accepted format is `--image-labels=key=val`.")
 	flag.Var(&containerImages, "container-image", "container image to include in the disk image. This flag can be specified multiple times")
 
+	instanceDiskSizeGb := newInstanceSizeGBFlag(diskSizeGb)
+	flag.Var(instanceDiskSizeGb, "instance-disk-size-gb", "disk size of the GCE instance used for disk image creation. If not set, defaults to the value of --disk-size-gb.")
```
elfinhe, line 70: Can help explain why we need the struct intInstanceSizeGBFlag?

This line and the struct can be replaced by
```
instanceDiskSizeGb := flag.Int64("instance-disk-size-gb", *diskSizeGb, "disk size of the GCE instance used for disk image creation. If not set, defaults to the value of --disk-size-gb.")
```

ctrlok: Actually I'm not sure right now. I remember that it was my first guess to use diskSizeGb — it's simple and didn't require additional code, but then it didn't work or I was unsatisfied with what I saw in `--help` menu

spencer-p:
Closing old issues with merge conflicts. Please rebase and re-open with tests if still relevant.

PR #358, opened 2024-03-14
Add source files and user guide for single host inferencing with JetStream and MaxText on GKE

vivianrwu:
This PR will provide a user guide to deploy a Gemma model with JetStream repo e2e on GKE


jetstream/maxtext/maxengine-server/Dockerfile
puviyarasan, line 1: Does it make sense for the containerization of maxengine-server to be in the MaxText repo? 
cc: @ahg-g

jetstream/maxtext/single-host-inference/README.md
```diff
@@ -0,0 +1,157 @@
...
+    --machine-type=ct5lp-hightpu-8t \
+    --project=${PROJECT_ID}
+```
+You have created the following resources:
+
+- Standard cluster with 2 CPU nodes.
+- One v5e TPU node pool with 2 nodes, each with 8 chips.
+
+### Configure Applications to use Workload Identity
+Prequisite: make sure you have the following roles
```
puviyarasan, line 40: typo: Pre-requisite

jetstream/maxtext/maxengine-server/Dockerfile
vivianrwu: Could you clarify on this? The maxengine-server server source code is in the MaxText repo `python3 MaxText/maxengine_server.py`, see https://github.com/google/maxtext/blob/main/MaxText/maxengine_server.py.

liurupeng: yep, I think we may need to add the dockerfile for containerization of maxengine to the Maxtext repo, so that anyone could use it as well

ahg-g: I think it is fine for now, but my main concern is that this fetches the main branch. We need to establish a release process for maxengine_server image with versions in the jetstream repo, and so ideally the dockerfile would also be released with it. A release is a named and frozen github branch (see https://github.com/kubernetes-sigs/lws/tree/release-0.1 for 0.1 release of lws as an example).

jetstream/maxtext/single-host-inference/deployment.yaml
```diff
@@ -0,0 +1,58 @@
...
+      containers:
+      - name: maxengine-server
+        image: us-docker.pkg.dev/cloud-tpu-images/inference/maxengine-server:v1.0.0
+        securityContext:
+          privileged: true
+        args:
+        - model_name=gemma-7b
+        - tokenizer_path=assets/tokenizer.gemma
+        - per_device_batch_size=1
+        - run_name=maxtext_jetstream_run
```
ahg-g, line 27: sorry to comment on this again, but this is still not clear to me and is sticking like a sore thumb :) does the value have meaning? is it associated with the checkpoint or the server binary? how do users discover it? or is it some arbitrary string?

jetstream/maxtext/single-host-inference/README.md
```diff
@@ -0,0 +1,157 @@
...
+    iam.gke.io/gcp-service-account=jetstream-iam-sa@${PROJECT_ID}.iam.gserviceaccount.com
+```
+
+## Deploy Maxengine Server and HTTP Server
+
+You can use the provided Maxengine server and HTTP server images already in `deployment.yaml` or [build your own](#optionals)
+
+Add desired overrides to your yaml file by editing the `args` in `deployment.yaml`. You can reference the [MaxText base config file](https://github.com/google/maxtext/blob/main/MaxText/configs/base.yml) on what values can be overridden.
+
+Use a checkpoint by adding `load_parameters_path=<GCS bucket path to your checkpoint>` under `args`
```
liurupeng, line 82: "Configure the model checkpoint by adding..."

liurupeng: "Or you could try deploy the sample yaml without adding the checkpoint path. In the `deployment.yaml`, we will deploy a Maxengine server targeting  Gemma-7b model" I feel we may add slightly more description about what model we are deploying and its characteristics so that user may understand the args easier, wdyt?

jetstream/maxtext/http-server/Dockerfile
achandrasekar, line 1: This one probably shouldn't be a top level directory since we are trying to reorganize it. @imreddy13 what is the plan here? It would be good to create the top level directories that we want to consolidate on, so new changes can go in there directly.

jetstream/maxtext/http-server/http_server.py
```diff
@@ -0,0 +1,90 @@
...
+import fastapi
+import grpc
+from jetstream.core.proto import jetstream_pb2
+from jetstream.core.proto import jetstream_pb2_grpc
+import pydantic
+
+
+class GenerateRequest(pydantic.BaseModel):
+  server: Optional[str] = "127.0.0.1"
+  port: Optional[str] = "9000"
```
achandrasekar, line 32: Should we default to port 8000 here too to keep it consistent?

jetstream/maxtext/http-server/http_server.py
```diff
@@ -0,0 +1,90 @@
...
+  session_cache: Optional[str] = ""
+  prompt: Optional[str] = "AB"
+  priority: Optional[int] = 0
+  max_tokens: Optional[int] = 3
```
achandrasekar, line 36: Should we add more reasonable defaults for sample prompt and max tokens? We should probably default max_tokens to 100 to see reasonable responses.

jetstream/maxtext/http-server/http_server.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+app = fastapi.FastAPI()
+executor = concurrent.futures.ThreadPoolExecutor(max_workers=200)
```
achandrasekar, line 40: How did we decide on 200 for the max workers? Also, how does it do under benchmarking when qps is over 200?

jetstream/maxtext/maxengine-server/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN git clone https://github.com/google/maxtext.git && \
+git clone https://github.com/google/JetStream.git
+
+RUN cd maxtext/ && \
+bash setup.sh
+
+RUN cd /JetStream && \
+pip install -e .
+
+RUN echo '#!/bin/bash\n\
```
achandrasekar, line 28: Would be good to create a separate entrypoint file with this script and copy it over instead of inlining since this is more error prone with spacing and future changes.

jetstream/maxtext/single-host-inference/deployment.yaml
```diff
@@ -0,0 +1,58 @@
...
+      nodeSelector:
+        cloud.google.com/gke-tpu-topology: 2x4
+        cloud.google.com/gke-tpu-accelerator: tpu-v5-lite-podslice
+      containers:
+      - name: maxengine-server
+        image: us-docker.pkg.dev/cloud-tpu-images/inference/maxengine-server:v1.0.0
+        securityContext:
+          privileged: true
+        args:
+        - model_name=gemma-7b
```
achandrasekar, line 24: Where are we getting the model checkpoint from?

jetstream/maxtext/http-server/Dockerfile
liurupeng: I saw many folders like jobset/pytorch etc is a top level directory, why this one doesn't qualify? We want to send out the guide before Next. If the plan of reorganizing is not finalized, what is the reasoning of blocking here

jetstream/maxtext/http-server/http_server.py
liurupeng: to be consistent with which one? this is the default one in Maxtext and jetStream repo

jetstream/maxtext/http-server/http_server.py
liurupeng: The implementation of the http server again is not for users' reference, it's highly unlikely in a real case, this will be used since jetsream already creates a grpc server

liurupeng: our benchmarking results is similar to CMCS's one, but what would be a reasonable value here though?

jetstream/maxtext/maxengine-server/Dockerfile
liurupeng: Could you elaborate why it's more error prone? thanks!

jetstream/maxtext/http-server/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# Ubuntu:22.04
+# Use Ubuntu 22.04 from Docker Hub.
+# https://hub.docker.com/_/ubuntu/tags?page=1&name=22.04
+ARG UBUNTU_VERSION=83f0c2a8d6f266d687d55b5cb1cb2201148eb7ac449e4202d9646b9083f1cee0
```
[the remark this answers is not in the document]
liurupeng: I don't think a specific version here matters too much, just we need to have some image to use. Users who use this guide could change it themselves.

jetstream/maxtext/http-server/http_server.py
vivianrwu: The maxengine server uses exposes 9000 so we have to communicate via 9000 as well - https://github.com/google/maxtext/blob/main/MaxText/maxengine_server.py#L46.

jetstream/maxtext/http-server/Dockerfile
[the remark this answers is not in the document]
vivianrwu: There's not really a specific reason, this was pulled from the saxml repo where they used this image as well - https://github.com/google/saxml/blob/main/saxml/tools/docker/Dockerfile.model. So more so reusing what was used before. Users can choose to change this version too or we can change to just use ubuntu:22.04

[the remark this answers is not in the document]
ahg-g: Thanks, ok, lets do that then, more readable.

jetstream/maxtext/http-server/http_server.py
liurupeng: what is the max_token means here? if we set it to 3, why we still see reasonable response with more than 100 tokens output?

jetstream/maxtext/single-host-inference/deployment.yaml
vivianrwu: Thanks for actively following up with this. Previously I was under the impression that it was needed (because in training it was used) and via some discussions internally. For inferencing, I just tested with omitting run_name and no errors arise so I will omit the run_name from the yaml

jetstream/maxtext/http-server/http_server.py
liurupeng: users could override it, that's point. If we do need to update all these, then prompt here should be "", per_device_batch_size should be 12

vivianrwu: Taking another look at this, the sample prompts are in the UG here - https://github.com/GoogleCloudPlatform/ai-on-gke/pull/358/files/75e621e3feb329d34c6502c97d81db97f6ba6f05#diff-e1f2497210c64d2163b6cf01a41229ecd8ed3c4f9304d0d46a6aadab69189e7eR127. These are just defaults if nothing is passed. Assuming that prompt and max tokens are not passed, wouldn't it make more sense to have the response to be expected to be not so good?

jetstream/maxtext/http-server/http_server.py
achandrasekar: If this is pointing to the GRPC port, that makes sense. I thought we had a different default for the http port here.

jetstream/maxtext/http-server/http_server.py
achandrasekar: For the prompt, something like "This is an example prompt" would make more sense because we want to make that clear. Max tokens to 100 so that when someone doesn't pass the max tokens, we will still provide a response of reasonable length instead of terminating at 3 words or less.

> users could override it, that's point

The point of the defaults is to set something that when not provided works reasonably well. Of course users can and will override it. But in the case where someone doesn't, we don't want the defaults to be totally not useful.

jetstream/maxtext/http-server/http_server.py
achandrasekar: If we benchmarked the http server in our benchmarking and it worked well enough for our max throughput case, then I'm ok with this value. In general, I want to make sure we choose this value with the understanding of what the limitations are and if it works for the majority of the use cases.

jetstream/maxtext/maxengine-server/Dockerfile
achandrasekar: It is easier to make a mistake with adding escape sequences like `\n` where we miss a \ or misplace it leading to issues. Indentation is another problem where it is possible to mess up the indentation when you echo into a file. It is much cleaner to define the script outside and copy it in the Dockerfile using `COPY`. It is also easier to extend in the future if we need more changes.

jetstream/maxtext/single-host-inference/deployment.yaml
achandrasekar: If this yaml wouldn't work without a checkpoint path, then it would be good to include it in the args here with a placeholder for the path `<GCS_model_checkpoint_path>` to make it explicit. Also, are there instructions on where they can get the `gemma-7b` checkpoint for JetStream or how they can convert it from another format?

liurupeng: I don't think we need to put the placeholder here, it will makes it the only thing to fail the deployment. ButI would agree we need to give a pointer about how to convert the checkpoint @vivianrwu

jetstream/maxtext/maxengine-server/Dockerfile
liurupeng: the points you mentioned are very common, any single character goes wrong could make things not work, for now, I don't think we need to spend the effort to do that

jetstream/maxtext/http-server/http_server.py
liurupeng: ok, let's update it to 100 and configure the per_device_batch_size as 12, thanks @vivianrwu

jetstream/maxtext/single-host-inference/deployment.yaml
vivianrwu: The yaml works without a adding the checkpoint path.

achandrasekar: How? Are the model weights baked into the container image or are we downloading from somewhere?

vivianrwu: Yaml works without the checkpoint path but the responses with be nonsensical without a checkpoint. The model is baked into the container image from the maxtext repo - https://github.com/google/maxtext/blob/main/MaxText/configs/models/gemma-7b.yml

liurupeng: again, the whole point for the user guide here is for guidance. We don't have a public checkpoint for use, so we should not include it in the yaml. This is clearly stated in the readme

vivianrwu: To clarify, I will add the directions to convert the checkpoint, which is running the script here - https://github.com/google/maxtext/blob/main/end_to_end/test_gemma.sh#L14. Right now, we are not providing a converted final checkpoint for users.

jetstream/maxtext/http-server/http_server.py
vivianrwu: Thanks Ashok, changing to your suggestion!

jetstream/maxtext/http-server/http_server.py
vivianrwu: Thanks for bringing this up Ashok. We're only done initial benchmarks right now so it's hard to determine. With Saxml, we found success with 1000. I'll change it here for now since I anticipate that the benchmarking will be similar between Saxml and Jetstream/MaxText. We can adjust the value further in the future as well as we see fit.

jetstream/maxtext/single-host-inference/deployment.yaml
liurupeng: Yep, thanks for the suggestion Ashok, we should update the checkpoint once we have one from CMCS later

jetstream/maxtext/maxengine-server/Dockerfile
liurupeng: Ok, I will make this as a potential enhancement, we could iterate in it later when have more cycles, thanks so much!

jetstream/maxtext/http-server/http_server.py
liurupeng: thanks for bring this up, yes, let's use 1000 just in case there is any issue.

jetstream/maxtext/http-server/http_server.py
liurupeng: yep, thanks for the suggestion

jetstream/maxtext/http-server/Dockerfile
liurupeng: okay, I just pinged the aieco team, actually I think we need to add more major subfolders for the inference team, thanks!

achandrasekar (approved):
Thanks for making the changes!

PR #359, opened 2024-03-15
Refactor IAP & Marketplace UI

umeshkumhar:
- Remove brand and add create_brand flag
- add domain field to take inputs
- fix SQL Instance deletion bug
- update marketplace metadata yamls with display text.


applications/jupyter/main.tf
```diff
@@ -137,8 +137,7 @@ module "jupyterhub" {
   k8s_backend_config_name  = var.k8s_backend_config_name
   k8s_backend_service_name = var.k8s_backend_service_name
   k8s_backend_service_port = var.k8s_backend_service_port
-  url_domain_addr          = var.url_domain_addr
-  url_domain_name          = var.url_domain_name
-  members_allowlist        = var.members_allowlist
+  domain                   = var.domain
+  members_allowlist        = split(",", var.members_allowlist)
```
imreddy13, line 141: Does this PR also fix the issue with the allowlist or is that tracked separately? @blackzlq @umeshkumhar

applications/jupyter/variables.tf
```diff
@@ -92,21 +93,15 @@ variable "k8s_backend_service_port" {
...
+  type        = bool
+  description = "Create Brand OAuth Screen"
+  default     = false
 }
 
-variable "url_domain_name" {
+variable "domain" {
   type        = string
-  description = "Name of the domain provided by the user. This var will only be used if url_domain_addr is not empty"
+  description = "Provide domain for ingress resource and ssl certificate. If it's empty, it will use nip.io wildcard dns"
```
imreddy13, line 104: Can we add a hyperlink to nip.io so users know what it means?

We also discussed taking this as user input instead, so making the field non optional and forcing the user to type "nip.io". Ok to do this in a followup PR.

applications/jupyter/workloads.tfvars
```diff
@@ -28,24 +28,25 @@ cluster_membership_id = "" # required for private cluster, defaults to `cluster_
 
 ## JupyterHub variables
 namespace                         = "ml"
-gcs_bucket                        = "gcs-bucket-dsfhfh"
 create_gcs_bucket                 = true
-workload_identity_service_account = "jupyter-service-account"
+gcs_bucket                        = "gcs-bucket-dsfhfh"
```
imreddy13, line 32: Can we change this to "gcs-bucket-<unique suffix"> and add a comment saying it needs to be unique across all projects?

applications/rag/frontend/main.tf
```diff
@@ -58,6 +57,9 @@ resource "kubernetes_service" "rag_frontend_service" {
   metadata {
     name      = "rag-frontend"
     namespace = var.namespace
+    annotations = var.add_auth ? {
+      "beta.cloud.google.com/backend-config" = "{\"default\": \"${var.k8s_backend_config_name}\"}"
```
imreddy13, line 61: is this a bug fix? does it work without this?

applications/rag/metadata.yaml
```diff
@@ -17,27 +17,15 @@ spec:
   content: {}
   interfaces:
     variables:
-      - name: goog_cm_deployment_name
-        description: The name of the deployment.
-        varType: string
```
imreddy13, line 22: Why delete the deployment name?

applications/rag/metadata.yaml
```diff
@@ -17,27 +17,15 @@ spec:
...
-        required: true
-      - name: add_auth
```
imreddy13, line 24: Is this unused since showIf doesn't work?

applications/rag/metadata.yaml
```diff
@@ -17,27 +17,15 @@ spec:
...
         varType: string
-        defaultValue: "false"
-      - name: brand
-        description: name of the brand if there isn't already on the project. If there is already a brand for your project, please leave it blank and empty
-        varType: string
-        defaultValue: ""
-      - name: client_id
-        description: Client ID used for enabling IAP
+        defaultValue: false
+      - name: cloudsql_instance
```
imreddy13, line 23: is it possible to prefix the instance name and bucket name with the IM deployment name? like we do for the other resources (like SAs)?

applications/rag/metadata.yaml
```diff
@@ -50,17 +38,17 @@ spec:
...
         varType: string
         required: true
+      - name: create_brand
+        description: Create Brand OAuth Screen
+        varType: bool
+        defaultValue: false
       - name: create_cluster
         varType: bool
-        required: true
+        defaultValue: false
```
imreddy13, line 47: What does default = false do? Uncheck the create cluster box in the UI?

applications/rag/metadata.yaml
```diff
@@ -72,36 +60,114 @@ spec:
...
+        varType: bool
+        defaultValue: true
       - name: enable_grafana_on_ray_dashboard
         description: Add option to enable or disable grafana for the ray dashboard. Enabling requires anonymous access.
         varType: bool
         defaultValue: false
+      - name: frontend_add_auth
+        varType: bool
+        defaultValue: false
+      - name: frontend_client_id
```
imreddy13, line 75: Do we need these if they're not exposed in the UI?

applications/rag/metadata.yaml
```diff
@@ -72,36 +60,114 @@ spec:
...
+      - name: jupyter_k8s_backend_service_name
+        description: Name of the Backend Service
         varType: string
         defaultValue: proxy-public
-      - name: k8s_ingress_name
+      - name: jupyter_k8s_backend_service_port
+        description: Name of the Backend Service Port
+        varType: number
+        defaultValue: 80
+      - name: jupyter_k8s_iap_secret_name
```
imreddy13, line 148: Do we need these if they're not exposed in the UI?

applications/rag/outputs.tf
```diff
@@ -17,12 +17,17 @@ output "jupyterhub_uri" {
...
+  value       = module.jupyterhub.jupyterhub_ip_address
+}
+
 output "jupyterhub_user" {
   description = "JupyterHub user is only required for standard authentication. Ignore, in case of IAP authentication"
   value       = module.jupyterhub.jupyterhub_user
 }
 
-output "jupyter_password" {
+output "jupyterhub_password" {
```
imreddy13, line 30: Can we please update the README.md file as well, if we change output names? Otherwise users won't know how to use the terraform.

modules/cloudsql/main.tf
```diff
@@ -47,9 +47,10 @@ resource "random_password" "pwd" {
 }
 
 resource "google_sql_user" "cloudsql_user" {
-  name     = var.db_user
-  instance = google_sql_database_instance.main.name
-  password = random_password.pwd.result
+  name            = var.db_user
+  instance        = google_sql_database_instance.main.name
+  password        = random_password.pwd.result
+  deletion_policy = "ABANDON"
```
imreddy13, line 53: TY for fixing! fyi @blackzlq @artemvmin this fixes the issue with cloud SQL instance deletion.
@umeshkumhar Can you leave a comment explaining why this was added so it doesn't get accidentally deleted in the future?

modules/iap/iap.tf
```diff
@@ -19,17 +19,18 @@ data "google_project" "project" {
 
 # Creates a "Brand", equivalent to the OAuth consent screen on Cloud console
 resource "google_iap_brand" "project_brand" {
-  count             = var.brand == "" ? 1 : 0
+  count             = var.create_brand ? 1 : 0
   support_email     = var.support_email
-  application_title = "${var.app_name}-Application"
+  application_title = "Web Application"
```
imreddy13, line 24: Why change this?

modules/iap/iap.tf
```diff
@@ -108,15 +108,10 @@ resource "helm_release" "iap" {
...
-    name      = var.k8s_backend_service_name
-    namespace = var.namespace
-  }
-  annotations = {
-    "beta.cloud.google.com/backend-config" = "{\"default\": \"${var.k8s_backend_config_name}\"}"
-  }
-  depends_on = [helm_release.iap]
-}
\ No newline at end of file
+## TODO: grant permission to specific backend_service
```
imreddy13, line 111: What does this todo mean? What doesn't work today?

applications/rag/metadata.yaml
umeshkumhar: this is renmaed ot jupyter_add_auth & frontend_add_auth

applications/rag/metadata.yaml
umeshkumhar: its there.. just formatted to alphabetical order

applications/rag/metadata.yaml
umeshkumhar: will do in next PR.

applications/rag/metadata.yaml
umeshkumhar: false will showing exisiting cluster list first.. and then user can select new.. if create_cluster is required.

applications/rag/metadata.yaml
umeshkumhar: they are hidden for now.. they are kept.. so they are in sync with variables..

applications/rag/metadata.yaml
umeshkumhar: Not exactly but the blueprint generates metadata.yaml from variables.tf.. so it is better to manage here.. so if blueprint is triggered again.. it will respect the settings already configured.

applications/rag/outputs.tf
umeshkumhar: Uodated readme

modules/iap/iap.tf
umeshkumhar: Brand can be only one per project.. Keeping a single application name doesnot make sense.. jo kept generic.. else we can take input for that.

applications/rag/frontend/main.tf
umeshkumhar: this is replacement of kubernetes_annotation block in iap modules..

applications/jupyter/workloads.tfvars
umeshkumhar: okay

applications/jupyter/variables.tf
umeshkumhar: will add in marketplace UI.. for workloads.tfvars it still valid.

modules/iap/iap.tf
```diff
@@ -108,15 +108,10 @@ resource "helm_release" "iap" {
...
 
-resource "kubernetes_annotations" "annotation" {
-  api_version = "v1"
-  kind        = "Service"
-  metadata {
-    name      = var.k8s_backend_service_name
-    namespace = var.namespace
-  }
-  annotations = {
-    "beta.cloud.google.com/backend-config" = "{\"default\": \"${var.k8s_backend_config_name}\"}"
```
blackzlq, line 119: This is needed! otherwise the redeploy will miss the annotation

blackzlq (dismissed):
annotation for the service is needed, please test the redeploy like modify the frontend image sha

applications/jupyter/main.tf
blackzlq: This PR I think is trying to fix that. But I think we should use https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/iap_web_backend_service_iam

modules/iap/iap.tf
```diff
@@ -108,15 +108,10 @@ resource "helm_release" "iap" {
...
-  }
-  depends_on = [helm_release.iap]
-}
\ No newline at end of file
+## TODO: grant permission to specific backend_service
+resource "google_project_iam_member" "project" {
```
blackzlq, line 112: We should use https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/iap_web_backend_service_iam to add the allowlist for each service.

applications/rag/frontend/main.tf
blackzlq: No, please move back to iap module, we need to onboard ray too.

applications/jupyter/main.tf
umeshkumhar: yeah will take in separate PR.. dont want to stretch this one.

modules/iap/iap.tf
umeshkumhar: yeah will take in seprate PR.

applications/rag/frontend/main.tf
umeshkumhar: ray already has from helm charts..
using kuberntes_annotations brings lot of diffs in terraform state.. after every apply there will be always some change around annotations..

modules/iap/iap.tf
umeshkumhar: moved into application respective services

applications/rag/metadata.yaml
imreddy13: Hm, but why? I think we want them to choose create cluster option if possible.

imreddy13 (approved):

modules/iap/iap.tf
blackzlq: Test change image and redeploy please.

umeshkumhar: tested working fine with IAP on frontend and jupyter

applications/rag/frontend/main.tf
blackzlq: Ok, then we should add annotation `
      "cloud.google.com/neg" = "{\"ingress\": true}"`
      
    this annotation will be auto added first time apply, did you try re-apply?

applications/rag/frontend/main.tf
```diff
@@ -89,6 +92,11 @@ resource "kubernetes_service" "rag_frontend_service" {
 
     type = var.add_auth ? "NodePort" : "ClusterIP"
   }
+  lifecycle {
+    ignore_changes = [
```
blackzlq, line 96: Oh I see this one, ignore the previous one

PR #360, opened 2024-03-15
Enable api and update image

blackzlq:
1. Enable api for dlp and nature language api
2. update frontend image sha


applications/rag/frontend/main.tf
```diff
@@ -15,6 +15,24 @@ data "google_project" "project" {
   project_id = var.project_id
 }
 
+# Enabled the DLP service
+resource "google_project_service" "dlp_api" {
```
imreddy13, line 19: Tested tf apply and tf destroy work as expected I hope?

imreddy13 (approved):

imreddy13 (dismissed):

applications/rag/frontend/main.tf
blackzlq: The job is using default SA, https://cloud.google.com/build/docs/automating-builds/create-manage-triggers which lack of the permission, I tested on my local. it works

blackzlq: We need to update the sa for e2e test, this code change is ok

imreddy13: Does that mean this will be broken for customer projects too?

blackzlq: Two cases:
1. if customer using terraform, they should use a sa like owner (or we should give a permission list what they need)
2. if they deploy through gcp, then we can create the sa for them with the certain permission, like what we are doing today.

blackzlq: After I add service admin permission it passed the test

bjornsen (approved):

PR #361, opened 2024-03-15
Replace kuberay-cluster chart variants with a single yaml

artemvmin:
Additionally, use the head for running jobs instead of spinning up a (head, worker) pair. I have not found Ray documentation recommending otherwise. Besides, the current worker is tiny in terms of CPU and we are not using the multi-GPU setup as far as I'm aware.

I tested the entire RAG application E2E on Autopilot. I also generated all permutations of ray cluster helm charts using `terraform plan` to made sure the image, resources, annotations, and node selectors match their prior values.


modules/kuberay-cluster/values.yaml
```diff
@@ -152,108 +143,11 @@ head:
         name: ray-logs
       - mountPath: /fluent-bit/etc/
         name: fluentbit-config
+
 worker:
```
imreddy13, line 147: I don't think we should do this. let's remember that customers that use this repo may use just Ray, without RAG - and probably need good examples for how to use workers.

We either need a RAG specific ray config, or an option to disable workers for RAG, or should just leave this as is for now.

imreddy13: @richardsliu wdyt?

artemvmin: Done.

bjornsen (approved):

modules/kuberay-cluster/variables.tf
```diff
@@ -43,7 +43,7 @@ variable "namespace" {
 
 variable "enable_tpu" {
   type        = bool
-  description = "Set to true to request TPUs in Ray cluster"
+  description = "Set to true to request TPUs in Ray cluster (not supported on GKE Autopilot)"
```
andrewsykim, line 46: cc @ryanaoleary @richardsliu

PR #365, opened 2024-03-18
Remove T4 nodepool for RAG

imreddy13:
- Switch Ray workers to L4s

### DIFF
--- a/modules/kuberay-cluster/main.tf
+++ b/modules/kuberay-cluster/main.tf
@@ -68,7 +68,7 @@ resource "helm_release" "ray-cluster" {
         "iam.gke.io/gke-metadata-server-enabled" : "true"
         } : var.enable_gpu ? {
         "iam.gke.io/gke-metadata-server-enabled" : "true"
-        "cloud.google.com/gke-accelerator" : "nvidia-tesla-t4"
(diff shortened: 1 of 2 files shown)


modules/kuberay-cluster/main.tf
```diff
@@ -68,7 +68,7 @@ resource "helm_release" "ray-cluster" {
         "iam.gke.io/gke-metadata-server-enabled" : "true"
         } : var.enable_gpu ? {
         "iam.gke.io/gke-metadata-server-enabled" : "true"
-        "cloud.google.com/gke-accelerator" : "nvidia-tesla-t4"
+        "cloud.google.com/gke-accelerator" : "nvidia-l4"
```
andrewsykim, line 71: We also discussed just running the Ray cluster on CPUs right? 

Have we tested if generating the embeddings with CPUs is fast enough?

imreddy13: @blackzlq played with this but we haven't benchmarked, next PR

bjornsen (approved):

imreddy13:
AP test is broken. standard test passed on a previous run (only change is TF linting). bypassing

PR #375, opened 2024-03-19
support additional labels for rag, ray and jupyter

andrewsykim:
Adds a new variable `additional_labels` for rag, ray and jupyter.

### DIFF
--- a/modules/jupyter/jupyter_config/config-selfauth-autopilot.yaml
+++ b/modules/jupyter/jupyter_config/config-selfauth-autopilot.yaml
@@ -26,7 +26,7 @@ hub:
     name: us-docker.pkg.dev/ai-on-gke/jupyterhub-authentication-class/jupyter-auth-class
     tag: latest
   labels:
-    ${indent(4, chomp(yamlencode(additional_labels)))}
+    ${indent(4, chomp(jsonencode(additional_labels)))}
   config:
     JupyterHub:
       authenticator_class: ${authenticator_class}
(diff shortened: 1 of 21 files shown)


modules/jupyter/jupyter_config/config-selfauth-autopilot.yaml
```diff
@@ -25,6 +25,8 @@ hub:
   image:
     name: us-docker.pkg.dev/ai-on-gke/jupyterhub-authentication-class/jupyter-auth-class
     tag: latest
+  labels:
+    ${indent(4, chomp(jsonencode(additional_labels)))}
```
andrewsykim, line 29: For some reason `yamlencode` was returning invalid YAML here but jsonecode is working fine. 

From the docs this is fine:

```
YAML is a superset of JSON, and so where possible we recommend generating JSON using [jsonencode](https://developer.hashicorp.com/terraform/language/functions/jsonencode) instead, even if a remote system supports YAML. JSON syntax is equivalent to flow-style YAML and Terraform can present detailed structural change information for JSON values in plans, whereas Terraform will treat block-style YAML just as a normal multi-line string. However, generating YAML may improve readability if the resulting value will be directly read or modified in the remote system by humans.
```

I've also validated this works in my environment and from CI

andrewsykim: oddly this only applies to the jupyter config, the ray values.yaml had no issue with yamlencode

imreddy13 (approved):

PR #380, opened 2024-03-19
Add Ray Cluster ResourceQuotas

bjornsen:
Add ResourceQuotas to Ray cluster to prevent DoS and resource exhaustion. Some error and networkpolicy fixups too.


applications/ray/main.tf
```diff
@@ -157,3 +154,17 @@ module "kuberay-cluster" {
   depends_on             = [module.gcs, module.kuberay-operator]
 }
 
+
+# Assign resource quotas to Ray namespace to ensure that they don't overutilize resources
```
umeshkumhar, line 158: Thinking of taking resource_quota out of modules from rag/ray/jupyter and define a new module.. that will be invoked during namespace creation module or after that. 

these quotas are assigned early and are not configured at app module level.. (also since its namespace level config, it make more sense outside of app modules)

WDYT?

bjornsen: I don't see what benefit this gives us. Ultimately, we're still implementing a ResourceQuota on a namespace in a particular application. By wrapping this resource in a module, we're simply adding an additional layer of indirection that requires more work to plumb through options. It made sense on the Ray module since my goal is to harden the Ray cluster but I was fine pulling it up to the Ray application to try to make progress.

modules/kuberay-cluster/main.tf
```diff
@@ -105,44 +105,26 @@ resource "kubernetes_network_policy" "kuberay-head-network-policy" {
     }
 
     ingress {
-      # Ray Client server
```
blackzlq, line 108: In this case, will it break out README to submit the job using port forward?

applications/ray/variables.tf
```diff
@@ -183,3 +183,28 @@ variable "goog_cm_deployment_name" {
   type    = string
   default = ""
 }
+
+# These default resource quotas are set intentionally high as an example that won't be limiting for most clusters.
+# Consult https://kubernetes.io/docs/concepts/policy/resource-quotas/ for additional quotas that may be set.
+variable "resource_quotas" {
```
blackzlq, line 189: What will be the default value if we don't set this quota. I think our project only have 16 GPU as the quota. Will 100 be too more?

bjornsen: This only acts as an upper limit to the number of GPUs Ray namespace Pods can request in their manifests. It won't interact with GCP resource quotas at all and is set intentionally high as an example that won't limit most workloads.

modules/kuberay-cluster/main.tf
bjornsen: `kubectl port-forward` always works with networkpolicy

PR #391, opened 2024-03-20
Single-Host TPU Webhook Fix

ryanaoleary:
This PR defaults the single-host Pod `TPU_WORKER_ID` to 0 to fix a bug that was incrementing the ID upon deletion and re-creation. This PR also updates the code to only inject `TPU_WORKER_HOSTNAMES` for multi-host podslices (previously an empty string would be injected). This PR was manually tested by creating both single-host and multi-host RayClusters using this template: https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.tpu-v4-singlehost.yaml and verifying that the env vars were set correctly even after multiple creations and deletions of the RayCluster.

### DIFF
--- a/applications/ray/kuberay-tpu-webhook/main.go
+++ b/applications/ray/kuberay-tpu-webhook/main.go
@@ -439,8 +439,10 @@ func mutatePod(admissionReview *admissionv1.AdmissionReview) (*admissionv1.Admis
 			container := containers[i]
 			if containerRequestingTPUs(container) {
 				path := fmt.Sprintf("/spec/containers/%d/env", i)
-				// inject TPU_WORKER_HOSTNAMES set during RayCluster interception
-				injectHostnames(sliceToHostnames[podSlice], path, container, &patches)
+				if isMultiHost {
+					// inject TPU_WORKER_HOSTNAMES set during RayCluster interception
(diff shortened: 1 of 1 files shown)


richardsliu (approved):

applications/ray/kuberay-tpu-webhook/main.go
```diff
@@ -413,10 +413,13 @@ func mutatePod(admissionReview *admissionv1.AdmissionReview) (*admissionv1.Admis
 		numOfHosts, _ := getNumTPUHostsFromTopology(topology, acceleratorType) // ignore error here because topology may not be set yet
 		replicaIndex := getReplicaIndex(clusterName)
 		podSlice := slice{clusterName, groupName, replicaIndex, numOfHosts}
-		tpuWorkerID := getNextWorkerID(podSlice, replicaIndex)
+		tpuWorkerID := 0 // defaults to 0 for single-host
```
andrewsykim, line 416: Any way to add a unit test?

ryanaoleary: I'm working on adding a unit test suite for the webhook that'll cover this, it should be out by the end of the week.

PR #393, opened 2024-03-20
fix broken node selector in kuberay-cluster module

andrewsykim:


modules/kuberay-cluster/main.tf
```diff
@@ -75,7 +75,7 @@ resource "helm_release" "ray-cluster" {
         "cloud.google.com/gke-tpu-topology" : "2x2x1"
         "cloud.google.com/gke-placement-group" : "tpu-pool"
         } : {
-        "iam.gke.io / gke-metadata-server-enabled" : "true"
+        "iam.gke.io/gke-metadata-server-enabled" : "true"
```
andrewsykim, line 78: I'm guessing this was missed from CI because we don't test CPU-only yet

imreddy13 (approved):

PR #398, opened 2024-03-20
Add private IP support for modules/cloudsql

pwschuurman:
This adds support for creating a CloudSQL instance with Private IP only.

1) Adds PSC configuration to the cloudsql module. This allocates a PSC IP range and connection.
2) Updates consumers (Ray worker via Jupyter notebook, RAG frontend) to use `private` IP type when connection to CloudSQL.

TODO: In order for the Frontend service to function correctly, `applications/rag/frontend/main.tf` must be updated to point to the correct container images that contain changes to `applications/rag/frontend/container/main.py` within this PR. This image should be pushed and the tag updated prior to submitting this PR.


modules/cloudsql/main.tf
```diff
@@ -21,39 +21,89 @@ resource "google_project_service" "project_service" {
...
-  depends_on = [google_sql_database_instance.main]
+data "google_compute_network" "main" {
+  name    = var.network_name
+  project = var.project_id
 }
 
-resource "random_password" "pwd" {
-  length  = 16
-  special = false
+resource "google_compute_global_address" "google-managed-services-range" {
```
umeshkumhar, line 41: Had a discussion with @imreddy13, the ideal to setup PSA is in infrastructure module. Can we move this PSA required resource blocks to infra module? 

And cloudsql module will always expect that given network will have PSA already enabled and configured. (Also mention same in readme and variable description.)

In infrastructure whenever create_network=true, always setup network with PSA access.. for existing network, lets skip setting PSA for phase 1 at least. 

Once phase 1 works good for all and things are pretty stable, then we can support `create_psa` flag that could configure PSA on existing network.

pwschuurman: > In infrastructure whenever create_network=true, always setup network with PSA access.. for existing network, lets skip setting PSA for phase 1 at least.

Sounds reasonable. Will update.

applications/rag/main.tf
```diff
@@ -35,8 +35,8 @@ module "infra" {
   autopilot_cluster = var.autopilot_cluster
   private_cluster   = var.private_cluster
   create_network    = false
-  network_name      = "default"
-  subnetwork_name   = "default"
+  network_name      = var.network_name
+  subnetwork_name   = var.network_name
```
umeshkumhar, line 39: create a new var for subnet as well..

applications/rag/main.tf
```diff
@@ -131,7 +131,9 @@ module "cloudsql" {
   instance_name = var.cloudsql_instance
   namespace     = var.kubernetes_namespace
   region        = local.cloudsql_instance_region
-  depends_on    = [module.namespace]
+  network_name  = var.network_name
+  // Depend on module.infra to ensure network is created
+  depends_on = [module.namespace, module.infra]
```
umeshkumhar, line 136: Not required.  as kubernetes provider will be configured after finishing the infra module.. so implicitly it depends on already.

applications/rag/variables.tf
```diff
@@ -345,3 +345,9 @@ variable "disable_ray_cluster_network_policy" {
   type        = bool
   default     = true
 }
+
+variable "network_name" {
+  description = "Network and subnetwork name"
```
umeshkumhar, line 350: create separate for subnetwork

applications/rag/main.tf
pwschuurman: Removed

applications/rag/variables.tf
pwschuurman: Added `subnetwork_name`

applications/rag/README.md
```diff
@@ -36,6 +36,17 @@ gcloud container clusters create-auto ${CLUSTER_NAME:?} \
   --release-channel rapid
 ```
 
+### Enable [Private Service Connect](https://cloud.google.com/vpc/docs/private-service-connect)
+
+Note: Skip this step when `create_network = true` is set.
```
imreddy13, line 41: I guess this is a little confusing since we don't surface create_network = true at the RAG level. Maybe we move this sentence to the /infrastructure README? And the RAG README always suggests checking that PSC is enabled for the default VPC (since that's what we use).

applications/rag/workloads.tfvars
```diff
@@ -14,8 +14,13 @@
 
 project_id = "<your project ID>"
 
+# GCP Network configuration
+create_network  = false
```
imreddy13, line 18: not sure we want to expose these.. then we also need to expose the CIDR range fields which makes things complex. How about just using default VPC for RAG and requiring the user to enable PSC?

wdyt @umeshkumhar

infrastructure/main.tf
```diff
@@ -57,11 +63,50 @@ module "custom-network" {
...
+resource "google_service_networking_connection" "private_service_access" {
+  count                   = var.create_network ? 1 : 0
+  network                 = local.network_self_link
+  service                 = "servicenetworking.googleapis.com"
+  reserved_peering_ranges = [google_compute_global_address.google-managed-services-range[0].name]
+  # This will enable a successful terraform destroy when destroying CloudSQL instances
+  deletion_policy = "ABANDON"
+}
+
+// module "private-service-access" {
```
imreddy13, line 93: we discussed just using this instead

pwschuurman: I kept this in as VPC deletion seems to work successfully. I was able to run `terraform apply`, go through the demo steps, and then run `terraform destroy` successfully.

applications/rag/workloads.tfvars
pwschuurman: >  then we also need to expose the CIDR range fields which makes things complex

What if we just update `infrastructure` to auto create subnets? Then we can defer to GCP defaults for subnets, and we don't need to worry about.

This also fixes a problem where the the GKE cluster is created in a different region from the CloudSQL instance. Today this wouldn't work with a `create_network = true`, since custom-network only creates a single subnet region.

imreddy13: I think that should work - does TF support auto creating subnets?

infrastructure/main.tf
imreddy13: ah, due to the ABANDON? makes sense!

imreddy13: maybe leave a comment explaining why we need the ABANDON and why we didn't use the PSC module, so it doesn't get accidentally deleted in the future

applications/rag/workloads.tfvars
pwschuurman: TF does, but the auto created subnets do not have "Private Google Access" enabled. I've updated the script to create CIDR sub-ranges dynamically from a provided CIDR range. This effectively gives us auto subnets, but with the ability to tune their parameters.

infrastructure/main.tf
pwschuurman: Updated the comment

infrastructure/main.tf
```diff
@@ -36,37 +36,83 @@ locals {
...
+//  # This will enable a successful terraform destroy when destroying CloudSQL instances
+//  deletion_policy = "ABANDON"
+// }
+
 locals {
-  network_name    = var.create_network ? module.custom-network[0].network_name : var.network_name
-  subnetwork_name = var.create_network ? module.custom-network[0].subnets_names[0] : var.subnetwork_name
-  region          = length(split("-", var.cluster_location)) == 2 ? var.cluster_location : ""
-  regional        = local.region != "" ? true : false
+  network_name      = var.network_name
```
umeshkumhar, line 105: why we removed this to normal vartype?

infrastructure/main.tf
```diff
@@ -36,37 +36,83 @@ locals {
...
-    {
-      subnet_name           = var.subnetwork_name
-      subnet_ip             = var.subnetwork_cidr
-      subnet_region         = var.subnetwork_region
-      subnet_private_access = var.subnetwork_private_access
-      description           = var.subnetwork_description
-    }
-  ]
-
+  subnets = [for subnet_region, subnet_ip in zipmap(var.subnetwork_regions, local.subnetwork_cidr_ips) : {
```
umeshkumhar, line 52: Actually, we can remove subnets all together and just enable auto-mode that will reduce the inputs.

infrastructure/main.tf
pwschuurman: Is there an advantage to using the output vs. the input variable? My understanding is that this resource just reflects the input to the output (through google_compute_network name argument: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_network), so I don't see a good reason to use a more complicated expression here.

infrastructure/main.tf
pwschuurman: Auto created subnets do not have "Private Google Access" enabled when created. When I tested this out using auto-mode network, pods that were created in the (Autopilot) GKE cluster were unable to access the internet, and "Private Google Access" was not enabled on any of the auto-created subnets.

According to the documentation, the hack is to enable Cloud NAT, which auto enalbes PGA: https://cloud.google.com/nat/docs/nat-product-interactions#interaction-pga. It seems like this should get enabled (within the cluster's region here: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/infrastructure/main.tf#L201), but as mentioned above I ran into that error testing. IMO I prefer the explicit PGA subnet creation to relying on the Cloud NAT side effect, as it's not immediately obvious that Cloud NAT has this effect.

applications/rag/main.tf
pwschuurman: Done

applications/rag/README.md
```diff
@@ -36,6 +36,17 @@ gcloud container clusters create-auto ${CLUSTER_NAME:?} \
   --release-channel rapid
 ```
 
+### Enable [Private Service Connect](https://cloud.google.com/vpc/docs/private-service-connect)
+
+Note: Skip this step if setting `create_network = true`
```
umeshkumhar, line 41: Or mention in terms of prerequisites for an existing network are.

applications/rag/workloads.tfvars
umeshkumhar: yes its still the same, we are now creating multiple subnets in multiple regions.

but we still need inputs cidr range, newbits, list of locations.. 

Rather to keep it simple, I would suggest to be it in single region for now. and allow inputs for cidr range and region.. and set them default..
https://github.com/GoogleCloudPlatform/ai-on-gke/blob/b8b72a9ded1d2e0b3cf587a9b393b2a9eca2296f/infrastructure/platform.tfvars#L24

we assume and also recommended that SQL instance and GKE lies in same region.

infrastructure/main.tf
```diff
@@ -36,37 +36,83 @@ locals {
   }
 }
 
+data "google_compute_network" "existing-network" {
```
umeshkumhar, line 39: lets keep existing network PSA enable out for now..

because now we dont know if PSA is already enabled for exisiting network and that will lead to apply failures.

We do PSA for all new network only on create_network=true

pwschuurman: This data stanza isn't enabling PSA, it's just a data reference to an existing resource. However it's only being used to reference "network_self_link", which is only used in the PSA generation (when create_network=true). So I think I can follow the same pattern as "cloud-nat" module below, and directly reference module.custom-network[0], rather than a local variable that is conditionally set. That will allow me to remove this data stanza.

pwschuurman: Actually, I'll keep this stanza. Having the `data` block in here validates that the existing network actually exists during `terraform apply` (by returning an error if it does not exist). This gives the user immediate feedback if they have misconfigured their Terraform.

applications/rag/README.md
pwschuurman: Done (added to header)

applications/rag/README.md
pwschuurman: Good idea. I moved this note to the `cloudsql` module.

applications/rag/main.tf
```diff
@@ -65,7 +66,8 @@ locals {
   frontend_default_uri    = "https://console.cloud.google.com/kubernetes/service/${var.cluster_location}/${var.cluster_name}/${var.kubernetes_namespace}/rag-frontend/overview?project=${var.project_id}"
   jupyterhub_default_uri  = "https://console.cloud.google.com/kubernetes/service/${var.cluster_location}/${var.cluster_name}/${var.kubernetes_namespace}/proxy-public/overview?project=${var.project_id}"
   ## if cloudsql_instance_region not specified, then default to cluster_location region
-  cloudsql_instance_region = var.cloudsql_instance_region != "" ? var.cloudsql_instance_region : (length(split("-", var.cluster_location)) == 2 ? var.cluster_location : join("-", slice(split("-", var.cluster_location), 0, 2)))
+  cluster_location_region  = (length(split("-", var.cluster_location)) == 2 ? var.cluster_location : join("-", slice(split("-", var.cluster_location), 0, 2)))
+  cloudsql_instance_region = var.cloudsql_instance_region != "" ? var.cloudsql_instance_region : local.cluster_location_region
```
imreddy13, line 70: Why does this data block need the cloud sql instance region?

infrastructure/main.tf
```diff
@@ -36,37 +36,83 @@ locals {
...
-  ]
-
+  subnets = [for subnet_region, subnet_ip in zipmap(var.subnetwork_regions, local.subnetwork_cidr_ips) : {
+    subnet_name           = var.subnetwork_name
+    subnet_ip             = subnet_ip
+    subnet_region         = subnet_region
+    subnet_private_access = var.subnetwork_private_access
+    description           = var.subnetwork_description
+  }]
   #secondary_ranges = var.network_secondary_ranges
```
imreddy13, line 59: do we need these?

infrastructure/variables.tf
```diff
@@ -37,13 +31,20 @@ variable "subnetwork_name" {
 }
 
 variable "subnetwork_cidr" {
-  type    = string
-  default = "10.128.0.0/20"
+  type        = string
+  default     = "10.128.0.0/16"
```
imreddy13, line 35: is this just random?

imreddy13 (approved):

infrastructure/main.tf
umeshkumhar: okay its always better to use module outputs.. as this helps terraform to prepare dependency graphs and prepare better apply sequence..

so rather than adding explicts `depends_on` on network module  is better to use module outputs that terraform considers as implicy dependency..

pwschuurman: Got it, thanks for the explanation. Changed this back, and removed the `depends_on`

applications/rag/main.tf
pwschuurman: This is a local variable block. `local.cloudsql_instance_region` is used elsewhere in this file

infrastructure/main.tf
pwschuurman: No we don't. Removed.

infrastructure/variables.tf
pwschuurman: This is based on the default block for auto subnet. Added a comment.

infrastructure/main.tf
umeshkumhar: yeah i forget to delete the comment

umeshkumhar: Sorry my other left in draft.

Let be it single region and single subnet only. This change will impact the existing scenarios which are already tested and validated. 

We can do this change after next release.. as we have code freeze by next week Tuesaday

@pwschuurman

applications/rag/workloads.tfvars
```diff
@@ -12,24 +12,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-project_id = "<your project ID>"
+project_id = "psch-gke-dev"
```
umeshkumhar, line 15: revert this

applications/rag/workloads.tfvars
```diff
@@ -12,24 +12,29 @@
...
+network_name    = "default"
+subnetwork_name = "default"
 
 ## this is required for terraform to connect to GKE master and deploy workloads
-create_cluster    = true # Create a GKE cluster in the default network.
+create_cluster    = true # Create a GKE cluster in the specified network.
 autopilot_cluster = true
-cluster_name      = "<cluster_name>"
-cluster_location  = "us-central1"
+cluster_name      = "rag-gke-cluster-psch"
```
umeshkumhar, line 25: revert

infrastructure/main.tf
pwschuurman: Sounds good. I'll port that complexity into a new PR.

I was thinking that as well, this PR started to get a bit too large.

applications/rag/workloads.tfvars
pwschuurman: Done

applications/rag/workloads.tfvars
pwschuurman: Done

umeshkumhar (approved):

PR #399, opened 2024-03-20
Ray Dashboard IAP

blackzlq:
1. Integrate IAP with Ray Dashboard

Tested-by: zlq


modules/kuberay-cluster/outputs.tf
```diff
@@ -17,4 +17,8 @@ output "ray_cluster_uri" {
     length(data.kubernetes_service.head-svc.status[0].load_balancer) > 0 ? (
       length(data.kubernetes_service.head-svc.status[0].load_balancer[0].ingress) > 0 ?
   "${data.kubernetes_service.head-svc.status[0].load_balancer[0].ingress[0].ip}" : "") : "") : ""
+}
+
+output "ray_dashboard_uri" {
```
umeshkumhar, line 22: this is not seprate then above one. - ray_cluster_uri .. merge this into above output

applications/rag/outputs.tf
```diff
@@ -42,3 +42,8 @@ output "frontend_ip_address" {
   description = "Frontend gloabl IP address"
   value       = module.frontend.frontend_ip_address
 }
+
+output "ray_dashboard_uri" {
```
umeshkumhar, line 46: same check output of already existing ray-cluster

modules/kuberay-cluster/outputs.tf
blackzlq: This is separate only for the dashboard, it can not be used to submit job

applications/rag/outputs.tf
blackzlq: This is separate only for the dashboard, it can not be used to submit job

PR #403, opened 2024-03-20
Fix for gmp error on Autopilot clusters

### LINKED ISSUE #402: Autopilot e2e tests are flaky due to GMP webook
Sometimes the Autopilot e2e tests fail because GMP webhook:
```
Error: Internal error occurred: failed calling webhook "default.podmonitorings.gmp-operator.gke-gmp-system.monitoring.googleapis.com": failed to call webhook: Post "https://gmp-operator.gke-gmp-system.svc:443/default/monitoring.googleapis.com/v1/podmonitorings?timeout=10s": No agent available

  with module.kuberay-monitoring[0].helm_release.gmp-engine,
  on ../../modules/kuberay-monitoring/main.tf line 16, in resource "helm_release" "gmp-engine":
  16: resource "helm_release" "gmp-engine" {
```

My guess is that this is because the Autopilot cluster has no nodes initially so the webhook it can't serve this request.

hsachdevah:
Create balloon pods in the 'create gke cluster' step & wait for gke-gmp-system resources to be ready before proceeding with further tests

Closes #402

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -64,8 +64,21 @@ steps:
         -var=cluster_location=$_REGION \
         -auto-approve -no-color -lock=false
         echo "pass" > /workspace/gke_cluster_result.txt
+
+        # Get kube config
+        gcloud container clusters get-credentials \
+        ml-$SHORT_SHA-$_PR_NUMBER-$_BUILD_ID-cluster \
+        --location $_REGION \
+        --project $PROJECT_ID 
+
+        # Create Baloon Pods to be used for AP clusters
+        kubectl apply -f tests/priority-class.yaml
+        kubectl apply -f tests/baloon-pods.yaml
+
+        kubectl wait --all pods -n gke-gmp-system --for=condition=Ready --timeout=300s
+
     dir: 'infrastructure/'
-    allowFailure: $_ALLOW_FAILURE
+    allowFailure: true
     waitFor: ['validate platform', 'validate ray', 'validate jupyterhub', validate rag]
 
   - id: 'test ray cluster'
@@ -75,12 +88,6 @@ steps:
       - '-c'
       - |
         set -e
-
-        # Get kube config
-        gcloud container clusters get-credentials \
-        ml-$SHORT_SHA-$_PR_NUMBER-$_BUILD_ID-cluster \
-        --location $_REGION \
-        --project $PROJECT_ID 
 
         cd /workspace/applications/ray/
         terraform apply \
@@ -103,7 +110,7 @@ steps:
         ray job submit \
         --address=http://127.0.0.1:8265 -- python -c "import ray; ray.init(); print(ray.cluster_resources())"
         echo "pass" > /workspace/ray_result.txt
-    allowFailure: $_ALLOW_FAILURE
+    allowFailure: true
     waitFor: ['create gke cluster']
 
   - id: 'cleanup ray cluster'
@@ -125,7 +132,7 @@ steps:
         -var=gcs_bucket=gke-aieco-ray-$SHORT_SHA-$_BUILD_ID \
         -auto-approve -no-color -lock=false
 
-    allowFailure: $_ALLOW_FAILURE
+    allowFailure: true
     waitFor: ['test ray cluster']
 
   - id: 'test jupyterhub'
@@ -158,7 +165,7 @@ steps:
         cd /workspace/modules/jupyter/tests
         python3 test_hub.py $(cat /workspace/jupyterhub_host_url.txt) $_AUTOPILOT_CLUSTER
         echo "pass" > /workspace/jupyterhub_test_result.txt
-    allowFailure: $_ALLOW_FAILURE
+    allowFailure: true
 
   - id: 'cleanup jupyterhub'
     name: 'gcr.io/$PROJECT_ID/terraform'
@@ -178,7 +185,7 @@ steps:
         -var=gcs_bucket=gke-aieco-jupyter-$SHORT_SHA-$_BUILD_ID \
         -auto-approve -no-color -lock=false
 
-    allowFailure: $_ALLOW_FAILURE
+    allowFailure: true
     waitFor: ['test jupyterhub']
 
   - id: 'test rag'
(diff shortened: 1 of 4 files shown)


infrastructure/tests/baloon-pods.yaml
```diff
@@ -0,0 +1,25 @@
...
+      pod: placeholder-pod
+  template:
+    metadata:
+      labels:
+        pod: placeholder-pod
+    spec:
+      priorityClassName: placeholder-priority
+      terminationGracePeriodSeconds: 0
+      containers:
+      - name: ubuntu
```
andrewsykim, line 18: Can we use a smaller image for this? I don't recall but `ubuntu` is a fairly bloated image for what we need to do here

infrastructure/tests/baloon-pods.yaml
```diff
@@ -0,0 +1,25 @@
...
+        image: ubuntu
+        command: ["sleep"]
+        args: ["infinity"]
+        resources:
+            requests:
+              cpu: 200m
```
andrewsykim, line 24: Can the CPU requests be really small (10m?) or not specfied at all?

andrewsykim: Ah, I guess if not specified it'll default to 500m in Autopilot. How about 10m or 25m?

infrastructure/tests/baloon-pods.yaml
```diff
@@ -0,0 +1,25 @@
...
+              memory: 250Mi
```
andrewsykim, line 25: Same comment here, use less memory? 25Mi?

cloudbuild.yaml
```diff
@@ -64,6 +64,19 @@ steps:
...
+
+        # Get kube config
+        gcloud container clusters get-credentials \
+        ml-$SHORT_SHA-$_PR_NUMBER-$_BUILD_ID-cluster \
+        --location $_REGION \
+        --project $PROJECT_ID 
+
+        # Create Baloon Pods to be used for AP clusters
+        kubectl apply -f tests/priority-class.yaml
+        kubectl apply -f tests/baloon-pods.yaml
```
andrewsykim, line 76: nit: balloon-pods.yaml

cloudbuild.yaml
```diff
@@ -64,6 +64,19 @@ steps:
...
         -auto-approve -no-color -lock=false
         echo "pass" > /workspace/gke_cluster_result.txt
+
+        # Get kube config
+        gcloud container clusters get-credentials \
+        ml-$SHORT_SHA-$_PR_NUMBER-$_BUILD_ID-cluster \
+        --location $_REGION \
+        --project $PROJECT_ID 
+
+        # Create Baloon Pods to be used for AP clusters
```
andrewsykim, line 74: Check _AUTOPILOT_CLUSTER first?

cloudbuild.yaml
hsachdevah: Fixed

infrastructure/tests/baloon-pods.yaml
hsachdevah: Fixed

infrastructure/tests/baloon-pods.yaml
hsachdevah: Fixed to 10m

infrastructure/tests/baloon-pods.yaml
hsachdevah: Switched to busybox

cloudbuild.yaml
hsachdevah: Well it can run on both. Changing the comment for now.

infrastructure/tests/baloon-pods.yaml
umeshkumhar: busybox or alpine will be also smaller in size

artemvmin:
This only fixes cloud run. We need a solution using terraform dependencies so that it also fixes local application deployments, such as RAG, which deploys the cluster via terraform.

andrewsykim:
@artemvmin agreed but I don't think it needs to block this PR. We need to resolve the flaky tests so people don't think their PRs are blocked (even though the pre submit is optional).

cloudbuild.yaml
```diff
@@ -64,6 +64,20 @@ steps:
...
         echo "pass" > /workspace/gke_cluster_result.txt
+
+        # Get kube config
+        gcloud container clusters get-credentials \
+        ml-$SHORT_SHA-$_PR_NUMBER-$_BUILD_ID-cluster \
+        --location $_REGION \
+        --project $PROJECT_ID 
+
+        # Create balloon pods
+        kubectl apply -f tests/priority-class.yaml
```
andrewsykim, line 75:
```bash
# Autopilot clusters can have 0 nodes, which can cause some issues with modules in applications/
# Create a dummy deployment in Autopilot to pre-provision nodes
if [[ $_AUTOPILOT_CLUSTER == "true" ]]; then
  kubectl apply -f tests/priority-class.yaml
  kubectl apply -f tests/dummy-deployment.yaml

  kubectl wait deployments dummy-deployment -n default --for=condition=Available --timeout=300s
  kubectl wait --all pods -n gke-gmp-system --for=condition=Ready --timeout=300s
fi
```

infrastructure/tests/priority-class.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: placeholder-priority
```
andrewsykim, line 4: Do we really need the priority class?

andrewsykim: Will this pre-empt the placeholder pods when the firsrt module is installed? If not I think we should remove it

andrewsykim (approved):
LGTM

(feel free to address other comments in a follow-up PR)

andrewsykim:
Sorry, I didn't realize @artemvmin already opened a PR for this fix https://github.com/GoogleCloudPlatform/ai-on-gke/pull/407

hsachdevah:
Waiting on #407 to avoid duplicate effort

TheSpiritXIII:
Side note, the GMP team (which I am part of) is working on a design doc and prioritizing a fix for this. We would have to backport the fix once it's made available on the GKE rapid channel so it will take a while to propagate through. We definitely appreciate any stopgap solutions like this for now. Thanks!

andrewsykim:
Thanks for the update @TheSpiritXIII 

I think we can close this PR since we have other workarounds merged

PR #407, opened 2024-03-20
Wait for gmp-operator rollout before applying kuberay podmonitorings (obsolete)

artemvmin:
`kuberay-operator` appears to be deployed faster than the `gmp-operator`, causing `kuberay-monitoring` to create a `PodMonitorings` object before the GMP webhook server is ready to validate it.

With the dependency in place, Terraform waits for the `gmp-operator` (external data source) to be rolled out prior to deploying `gmp-engine`:

```
Plan: 2 to add, 0 to change, 0 to destroy.
module.kuberay-monitoring.null_resource.wait_for_gmp: Creating...
module.kuberay-monitoring.null_resource.wait_for_gmp: Provisioning with 'local-exec'...
module.kuberay-monitoring.null_resource.wait_for_gmp (local-exec): Executing: ["/bin/sh" "-c" "while ! kubectl rollout status -n gke-gmp-system deploy/gmp-operator; do sleep 5; done"]
module.kuberay-monitoring.null_resource.wait_for_gmp (local-exec): deployment "gmp-operator" successfully rolled out
module.kuberay-monitoring.null_resource.wait_for_gmp: Creation complete after 0s [id=6180172337812646967]
module.kuberay-monitoring.helm_release.gmp-engine: Creating...
module.kuberay-monitoring.helm_release.gmp-engine: Creation complete after 6s [id=gmp-engine]
```

Fixes https://github.com/GoogleCloudPlatform/ai-on-gke/issues/402

### DIFF
--- a/modules/kuberay-monitoring/main.tf
+++ b/modules/kuberay-monitoring/main.tf
@@ -12,14 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# google managed prometheus engine
+locals {
+  gmp_namespace = var.autopilot_cluster ? "gke-gmp-system" : "gmp-system"
+}
+
+# Temporary workaround to ensure the GMP webhook is installed before applying PodMonitorings.
+resource "null_resource" "wait_for_gmp" {
+  provisioner "local-exec" {
+    command = "while ! kubectl get cm -n ${local.gmp_namespace} collector &>/dev/null; do echo 'Waiting for gmp-operator...'; sleep 10; done"
+  }
+}
+
 resource "helm_release" "gmp-engine" {
   name             = "gmp-engine"
   chart            = "${path.module}/charts/gmp-engine/"
   namespace        = var.namespace
   create_namespace = var.create_namespace
-  # timeout increased to support autopilot scaling resources, and give enough time to complete the deployment 
-  timeout = 1200
+  timeout          = 600
   set {
     name  = "projectID"
     value = var.project_id
@@ -29,9 +38,10 @@ resource "helm_release" "gmp-engine" {
     name  = "serviceAccount"
     value = var.k8s_service_account
   }
+
+  depends_on = [null_resource.wait_for_gmp]
 }
 
-# grafana
 resource "helm_release" "grafana" {
   count            = var.enable_grafana_on_ray_dashboard ? 1 : 0
   name             = "grafana"
--- a/applications/rag/main.tf
+++ b/applications/rag/main.tf
@@ -201,6 +201,7 @@ module "kuberay-monitoring" {
   source                          = "../../modules/kuberay-monitoring"
   providers                       = { helm = helm.rag, kubernetes = kubernetes.rag }
   project_id                      = var.project_id
+  autopilot_cluster               = local.enable_autopilot
   namespace                       = var.kubernetes_namespace
   create_namespace                = true
   enable_grafana_on_ray_dashboard = var.enable_grafana_on_ray_dashboard
--- a/applications/ray/README.md
+++ b/applications/ray/README.md
@@ -6,6 +6,7 @@ See the [Ray on GKE](/ray-on-gke/) directory to see additional guides and refere
 ## Installation
 
 Preinstall the following on your computer:
+* Kubectl
 * Terraform
 * Gcloud
 
--- a/applications/ray/main.tf
+++ b/applications/ray/main.tf
@@ -127,6 +127,7 @@ module "kuberay-monitoring" {
   source                          = "../../modules/kuberay-monitoring"
   providers                       = { helm = helm.ray, kubernetes = kubernetes.ray }
   project_id                      = var.project_id
+  autopilot_cluster               = local.enable_autopilot
   namespace                       = var.kubernetes_namespace
   create_namespace                = true
   enable_grafana_on_ray_dashboard = var.enable_grafana_on_ray_dashboard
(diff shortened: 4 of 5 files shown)


imreddy13 (approved):

bjornsen (approved):

modules/kuberay-monitoring/main.tf
```diff
@@ -12,14 +12,24 @@
...
-# google managed prometheus engine
+locals {
+  gmp_namespace = var.autopilot_cluster ? "gke-gmp-system" : "gmp-system"
+}
+
+# Temporary workaround to ensure the GMP webhook is installed before applying PodMonitorings.
+# This dependency can be removed once the GMP webhook is moved from node to KCP.
+resource "null_resource" "wait_for_gmp" {
+  provisioner "local-exec" {
+    command = "while ! kubectl rollout status -n ${local.gmp_namespace} deploy/gmp-operator; do sleep 5; done"
```
andrewsykim, line 23: This might be slightly easier to read:
```
kubectl wait --all pods -n gke-gmp-system --for=condition=Ready --timeout=300s
```

andrewsykim, line 23: Does this assume kubectl is available locally? Is that always true?

modules/kuberay-monitoring/main.tf
```diff
@@ -12,14 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# google managed prometheus engine
+# Temporary workaround to ensure the GMP webhook is installed before applying PodMonitorings.
+# This dependency can be removed once the GMP webhook is moved from node to KCP.
```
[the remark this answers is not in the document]
andrewsykim: (the first line is fine, specifically talking about the second line)

modules/kuberay-monitoring/main.tf
```diff
@@ -12,14 +12,20 @@
...
+resource "null_resource" "wait_for_gmp" {
+  provisioner "local-exec" {
+    command = "while ! kubectl rollout status -n gke-gmp-system deploy/gmp-operator; do sleep 5; done"
```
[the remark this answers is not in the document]
artemvmin: Yes. That should be a requirement for ai-on-gke demo? I know RAG mentions it explicitly. I also added it to the Ray README.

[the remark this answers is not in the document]
andrewsykim: I'm specifically thinking about scenarios where the environment running `terraform apply` is different from the user environment. @imreddy13 do you know if the the envionrment deploying the blueprints has kubectl availabe or is it just terraform?

modules/kuberay-monitoring/main.tf
artemvmin: I actually changed it to look for the presence of a configmap produced by the deployment. This should be a more reliable indicator and reduce flakiness.

Unfortunately, `kubectl wait` only works on created objects.

artemvmin: (Closing duplicate comment)

modules/kuberay-monitoring/main.tf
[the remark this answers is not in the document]
artemvmin: Done.

modules/kuberay-monitoring/main.tf
artemvmin: `kubectl watch` also wouldn't work in the case that `gmp-operator` has not yet been applied (race condition). The `null_resource` would never be created in that case.

modules/kuberay-monitoring/main.tf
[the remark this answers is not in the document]
artemvmin: Another option is to create a new terrraform provider to wait on a k8s resource.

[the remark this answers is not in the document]
umeshkumhar: This is not a recommended way to go forward.. Earlier also we refactored all local and remote provisioners..

Also this needs to be confirmed on marketplace, as marketplace execution environment may not have kubectl binary available and will lead to failures.

modules/kuberay-monitoring/main.tf
umeshkumhar: @artemvmin  if #403  works (we tested manually @hsachdevah  and it worked ), then maybe we can just include a dummy deployment that would trigger nodes, and also gmp-operator stables during that.. after that actual application deployment would not show any flakiness.

andrewsykim: @umeshkumhar I think the concern is that #403 is only fixing the CI tests, but it doesn't actually resolve the flake when actually deploying the terraform module. 

If we take the dummy deployment approach, it needs to be included in every Autopilot deployment, not just in the tests. But personally I would like to avoid dummy deployment if possible

artemvmin: I have an idea for a provider-based solution that doesn't use kubectl. I'll work on that today.

artemvmin:
Replaced by https://github.com/GoogleCloudPlatform/ai-on-gke/pull/418

modules/kuberay-monitoring/main.tf
artemvmin: The provider solution was too involved considering we have a GMP solution coming down the pike. Opting for a simple delay.

modules/kuberay-monitoring/main.tf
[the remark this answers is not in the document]
artemvmin: Agreed. Closing this PR.

PR #413, opened 2024-03-21
Network policy cleanup

bjornsen:
* Add Ray client port (10001)
* Only apply external ingress ports to Ray head

### DIFF
--- a/modules/kuberay-cluster/main.tf
+++ b/modules/kuberay-cluster/main.tf
@@ -95,7 +95,7 @@ data "kubernetes_service" "head-svc" {
 }
 
 # Allow ingress to the kuberay head from outside the cluster
-resource "kubernetes_network_policy" "ray-job-and-dashboard-namespace-network-policy" {
+resource "kubernetes_network_policy" "kuberay-head-namespace-network-policy" {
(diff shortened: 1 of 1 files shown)


modules/kuberay-cluster/main.tf
```diff
@@ -116,6 +116,12 @@ resource "kubernetes_network_policy" "ray-job-and-dashboard-namespace-network-po
         protocol = "TCP"
       }
 
+      # Ray client
+      ports {
```
umeshkumhar, line 120: do we need to include other ray cluster port as well?  metrics port for sure.. so gmp pods can poll metrics from this endpoint.

bjornsen: All ports are allowed within the namespace and for the gke-gmp-system namespace under the `kuberay-cluster-allow-network-policy` policy below. I don't want to make this too granular since it's only an example.

bjornsen: The intention of these firewall rules is to be configurable for end user to access the dashboard, client, and job submission ports.

umeshkumhar (approved):

PR #414, opened 2024-03-21
fix marketplace description & sql name

umeshkumhar:
- fix marketplace description
- fix sql instance name with prefix
- Rename Jupyterhub to JupyterHub

### DIFF
--- a/applications/jupyter/metadata.display.yaml
+++ b/applications/jupyter/metadata.display.yaml
@@ -178,11 +178,11 @@ spec:
         - heading: "Step 1: Create DNS A Records for Jupyterhub"
           description: |-
             If using custom domains for Jupyterhub, create DNS A record set (<a href="https://cloud.google.com/dns/docs/records#add_a_record">Google DNS Record Set</a>). Propagation takes 10-15 minutes and logging in won’t succeed until it’s done.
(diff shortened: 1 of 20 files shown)


applications/jupyter/metadata.display.yaml
```diff
@@ -178,11 +178,11 @@ spec:
         - heading: "Step 1: Create DNS A Records for Jupyterhub"
           description: |-
             If using custom domains for Jupyterhub, create DNS A record set (<a href="https://cloud.google.com/dns/docs/records#add_a_record">Google DNS Record Set</a>). Propagation takes 10-15 minutes and logging in won’t succeed until it’s done.
-        - heading: "Step 2: Go to JupyterHub Application"
+        - heading: "Step 2: Go to JupyterHub Application with sample notebooks"
```
imreddy13, line 181: either change it to: Go to Jupyterhub application and execute sample notebooks" or just leave it as is (in the heading)

imreddy13, line 181: Can we not do camel case within the word (so Jupyterhub instead of JupyterHub - lowercase h)

imreddy13 (approved):

PR #419, opened 2024-03-21
bump frontend image package version

yiyinglovecoding:
bump some package versions to decrease image vulnerabilities


imreddy13 (approved):
Will we update the image tag in the next PR (once Louhi pushes the image)?

imreddy13 (approved):

PR #420, opened 2024-03-21
Generate TPU name based on multihost group index

richardsliu:


applications/ray/kuberay-tpu-webhook/main.go
```diff
@@ -464,7 +464,7 @@ func mutatePod(admissionReview *admissionv1.AdmissionReview) (*admissionv1.Admis
 				if getEnvironmentVariable("TPU_WORKER_NAME", container) == "" {
```
ryanaoleary, line 464: I think this should actually be changed to `TPU_NAME`, rather than `TPU_WORKER_NAME`, otherwise LGTM.

ryanaoleary (approved):

applications/ray/kuberay-tpu-webhook/main.go
```diff
@@ -464,7 +464,7 @@ func mutatePod(admissionReview *admissionv1.AdmissionReview) (*admissionv1.Admis
...
 					tpuName := corev1.EnvVar{
 						Name:  "TPU_NAME",
-						Value: fmt.Sprint(groupName),
+						Value: fmt.Sprint("%s-%d", groupName, replicaIndex),
```
ryanaoleary, line 467: This should also be `Sprintf` instead of `Sprint` I believe.

PR #427, opened 2024-03-22
tpu-provisioner - Dedup NP creations on job-key and reduce noisy node pool deletion reconciles

nstogner:
Rapid JobSet restarts can trigger concurrent node pool creations for the same logical Job. This PR adds in-memory de-duplication of those scenarios.

Also: Reduce some noisy reconciles that occur when node pools are deleting.

### DIFF
--- a/tpu-provisioner/internal/cloud/gke.go
+++ b/tpu-provisioner/internal/cloud/gke.go
@@ -26,6 +26,7 @@ const (
 	GKEAcceleratorNodeSelector = "cloud.google.com/gke-tpu-accelerator"
 	GKENodePoolNameLabel       = "cloud.google.com/gke-nodepool"
 	GKENodePoolNamePrefix      = "tpu-provisioner-"
+	jobKeyLabel                = "jobset.sigs.k8s.io/job-key"
 	V4PodSliceAccelerator      = "tpu-v4-podslice"
 	V5ePodSliceAccelerator     = "tpu-v5-lite-podslice"
 	V5pPodSliceAccelerator     = "tpu-v5p-slice"
@@ -48,8 +49,9 @@ type GKE struct {
 
 	Recorder record.EventRecorder
 
-	inProgressDeletes sync.Map
-	inProgressCreates sync.Map
+	inProgressDeletes       sync.Map
+	inProgressCreates       sync.Map
+	inProgressCreatesJobKey sync.Map
 }
 
 func (g *GKE) NodePoolLabelKey() string { return GKENodePoolNameLabel }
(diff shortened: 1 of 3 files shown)


tpu-provisioner/internal/cloud/gke.go
```diff
@@ -80,11 +82,21 @@ func (g *GKE) EnsureNodePoolForPod(p *corev1.Pod, why string) error {
...
 	if _, inProgress := g.inProgressCreates.Load(name); inProgress {
-		return ErrDuplicateRequest
+		return fmt.Errorf("creation ongoing for node pool name: %v: %w", name, ErrDuplicateRequest)
 	}
 	g.inProgressCreates.Store(name, struct{}{})
 	defer g.inProgressCreates.Delete(name)
 
+	// A restarting JobSet will trigger a new Node Pool creation.
+	// This creation might overlap with the previous one, so we need to deduplicate.
+	if jobKey := p.Labels[jobKeyLabel]; jobKey != "" {
```
danielvegamyhre, line 92: Can we add to the comment here, explaining the jobKey is stable through restarts but the job UID (used for NP name) is not, so that is why we need to dedupe on the job key

tpu-provisioner/internal/cloud/gke.go
```diff
@@ -48,8 +49,9 @@ type GKE struct {
 
 	Recorder record.EventRecorder
 
-	inProgressDeletes sync.Map
-	inProgressCreates sync.Map
+	inProgressDeletes       sync.Map
```
danielvegamyhre, line 52: Can we either update the variable name or add a comment so that it's clear `inProgressDeletes` and `inProgressCreates` store node pool names to do the tracking?

danielvegamyhre:
LGTM overall, left a couple minor comments

nstogner:
Good comments, will make those changes

danielvegamyhre (approved):

tpu-provisioner/internal/cloud/gke.go
```diff
@@ -311,6 +326,13 @@ func sumTPURequests(p *corev1.Pod) (int, error) {
 }
 
 func podToNodePoolName(p *corev1.Pod, prefix, suffix string) string {
+	// Use the UID of the Pod as the unique identifier for the node pool.
```
danielvegamyhre, line 329: Nit: I think this should say we try to use the UID of the Job which owns the pod first, but only fall back to the pod UID if owner UID is unavailable, right?

nstogner: Good point

thebinaryone1 (approved):

PR #429, opened 2024-03-22
 Add before you begin checkbox for marketplace

umeshkumhar:
- update marketplace region list for ray
- update before you begin checkbox for marketplace.


modules/kuberay-operator/kuberay.tf
```diff
@@ -17,7 +17,7 @@ resource "helm_release" "kuberay-operator" {
   repository       = "https://ray-project.github.io/kuberay-helm/"
   chart            = "kuberay-operator"
   values           = var.autopilot_cluster ? [file("${path.module}/kuberay-operator-autopilot-values.yaml")] : [file("${path.module}/kuberay-operator-values.yaml")]
-  version          = "1.0.0"
+  version          = "1.1.0"
```
imreddy13, line 20: do we need to do this before Next? let's not

umeshkumhar: its just helm chart version. that is yamls file changes.. whcih has some k8s role permisson changes.. Doesnot seems it will impact application.

also tested e2e rag with sample jupyter dataset

imreddy13 (approved):

PR #432, opened 2024-03-22
Catch Frontend DatabaseError

yiyinglovecoding:
Frontend will not break when 
- Database not generated 
- Database return empty result

### DIFF
--- a/applications/rag/frontend/container/main.py
+++ b/applications/rag/frontend/container/main.py
@@ -140,21 +140,27 @@ def index():
 
 def fetchContext(query_text):
     with db.connect() as conn:
-        results = conn.execute(sqlalchemy.text("SELECT * FROM " + TABLE_NAME)).fetchall()
-        log.info(f"query database results:")
-        for row in results:
-            print(row)
+        try: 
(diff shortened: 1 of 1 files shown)


applications/rag/frontend/container/main.py
```diff
@@ -140,20 +140,28 @@ def index():
...
+
+            # chunkify query & fetch matches
+            query_emb = transformer.encode(query_text).tolist()
+            query_request = "SELECT id, text, text_embedding, 1 - ('[" + ",".join(map(str, query_emb)) + "]' <=> text_embedding) AS cosine_similarity FROM " + TABLE_NAME + " ORDER BY cosine_similarity DESC LIMIT 5;"
+            query_results = conn.execute(sqlalchemy.text(query_request)).fetchall()
+            conn.commit()
+            log.info(f"printing matches:")
+            for row in query_results:
+                print(row)
+        except sqlalchemy.exc.DatabaseError as err:
```
imreddy13, line 157: I wonder if we should make this more generic - catch all other Exceptions too and return empty string?

yiyinglovecoding: done

imreddy13 (approved):

Issue #428, opened 2024-03-22
JupyterHub service account is missing GCS related roles


andrewsykim:
When I tried to run JupyterHub backed by a GCSFuse, I ran into this error from JupyterHub:
```
2024-03-22T18:48:02Z [Warning] MountVolume.SetUp failed for volume "gcs-fuse-csi-ephemeral" : rpc error: code = PermissionDenied desc = failed to get GCS bucket "gcsfuse-admin": googleapi: Error 403: jupyter-sa@<project-id>.iam.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket. Permission 'storage.objects.list' denied on resource (or it may not exist)., forbidden
```

It looks like the predefined roles are missing GCS related roles: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/5a80eb2203bff0cbc53d31292516a62e12ca3abc/modules/jupyter/variables.tf#L54-L59

andrewsykim:
@chiayi can you take a look? cc @imreddy13

PR #434, opened 2024-03-23
JupyterHub should use a GCS bucket it can access when using GCSFuse

andrewsykim:
Fixes https://github.com/GoogleCloudPlatform/ai-on-gke/issues/428

When using GCSFuse to back notebook data, JupyerHub is configured to use buckets named `gcsfuse-{username}`. This doesn't work because the bucket with that name is not created as part of any of our guides and the google service account does not have permissions to access those buckets. The buckets may also not be available because they're not unique.

This PR updates the ephemeral GCSFuse mounts to use the same GCS bucket created in the terraform module. It also configures the GCSFuse mount to use a subdirectory in the bucket `notebooks/{username}`.

### DIFF
--- a/modules/jupyter/jupyter_config/config-selfauth-autopilot.yaml
+++ b/modules/jupyter/jupyter_config/config-selfauth-autopilot.yaml
@@ -135,8 +135,8 @@ singleuser:
                   csi:
                     driver: gcsfuse.csi.storage.gke.io
                     volumeAttributes:
-                      bucketName: gcsfuse-{username}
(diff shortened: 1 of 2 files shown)


chiayi:
LGTM, but just wondering if it mounts the same bucket twice? Once as an `extra_volume` and another as the normal storage solution.

chiayi (approved):

andrewsykim:
> LGTM, but just wondering if it mounts the same bucket twice? Once as an extra_volume and another as the normal storage solution.

That's correct, but it mounts the bucket at different paths. One in `/data` and the others in `/notebooks/{username}`

PR #437, opened 2024-03-25
Update frontend image

blackzlq:
consume the latest change

Tested-by: No test


imreddy13 (approved):

yiyinglovecoding:
@imreddy13  do we want to replace the hash with image tag instead?

blackzlq:
> @imreddy13 do we want to replace the hash with image tag instead?

No, we prefer SHA than tag for the jupyter case we can not use SHA.

imreddy13 (approved):

PR #441, opened 2024-03-25
Update build to use us-east4 for L4 capacity

artemvmin:
Also bump the GMP delay


artemvmin:
Sounds like us-east4 is even more constrained.

PR #449, opened 2024-03-26
use single namespace installation for kuberay-operator

andrewsykim:
So that kuberay resources are scoped to only the namespaced used in the terraform deployment


imreddy13 (approved):
Think I prefer this one if it works, but if testing is a pain, 448 is ok too!

PR #462, opened 2024-03-27
support output namespace field

umeshkumhar:
- expose namespace name as terraform ouput and marketplace outputs.


applications/jupyter/metadata.yaml
```diff
@@ -152,3 +152,5 @@ spec:
         description: JupyterHub Endpoint to access user interface. In case of private IP, consider port-forwarding.
       - name: jupyterhub_user
         description: JupyterHub user is only required for standard authentication. Ignore, in case of IAP authentication
+      - name: kubernetes_namespace
```
andrewsykim, line 155: Please add description in a follow-up

andrewsykim (approved):

PR #464, opened 2024-03-27
Update README.md

jsoohoo-google:
Include servicenetworking.googleapis.com as dependency

### DIFF
--- a/infrastructure/README.md
+++ b/infrastructure/README.md
@@ -17,6 +17,7 @@ For the GCP project where the infra resources are being created, the following p
 Following service APIs are enabled, 
 - container.googleapis.com
 - gkehub.googleapis.com
+- servicenetworking.googleapis.com
 
 if not already enabled, use the following command:
 ```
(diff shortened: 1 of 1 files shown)


infrastructure/README.md
```diff
@@ -17,6 +17,7 @@ For the GCP project where the infra resources are being created, the following p
 Following service APIs are enabled, 
 - container.googleapis.com
 - gkehub.googleapis.com
+- servicenetworking.googleapis.com
```
andrewsykim, line 20: Also add cloudresourcemanager.googleapis.com please

andrewsykim: @umeshkumhar can we consider enabling these automatically in the `infrastructure` module, similar to what we do with `rag`, `ray,` and `jupyter`?

andrewsykim (approved):

infrastructure/README.md
andrewsykim: https://github.com/GoogleCloudPlatform/ai-on-gke/issues/474

PR #466, opened 2024-03-27
Downgrade RAG to 1.28; bump GMP resource creation delay; increase Ray worker creation timeout

artemvmin:
We are still seeing flakiness.

Pip installs are experiencing intermittent failures. Downgrading to 1.28 to rule out possibility of k8s or GKE regressions.

For the GMP operator, one explanation is that there are multiple system pods all starting up simultaneously and competing for resources, delaying `gmp-operator` startup. Bumping this to 60s to rule out this issue entirely. This shouldn't affect E2E creation, because the long tail is still CloudSQL instance creation.

Context: https://github.com/GoogleCloudPlatform/ai-on-gke/pull/418

Additionally, increase the Ray worker timeout to 1200s (from 300s). Although, for some reason the `kuberay-cluster` helm chart release is not waiting for all resources to be ready:

```
...
module.kuberay-cluster.helm_release.ray-cluster: Creation complete after 2s
...
error: timed out waiting for the condition on pods/ray-cluster-kuberay-worker-workergroup-8ljp7
...
```

It can't possibly be provisioning a GPU node for the worker in 2 seconds. Adding an explicit `wait` to rule out a defaulting issue.

This is an open issue since 2021: https://github.com/hashicorp/terraform-provider-helm/issues/672

### DIFF
--- a/applications/rag/variables.tf
+++ b/applications/rag/variables.tf
@@ -31,6 +31,11 @@ variable "cluster_location" {
   type = string
 }
 
+variable "kubernetes_version" {
+  type    = string
+  default = "1.28"
+}
+
 variable "kubernetes_namespace" {
   type        = string
   description = "Kubernetes namespace where resources are deployed"
--- a/modules/kuberay-monitoring/main.tf
+++ b/modules/kuberay-monitoring/main.tf
@@ -14,7 +14,8 @@
 
 # Temporary workaround to ensure the GMP webhook is installed before applying PodMonitorings.
 resource "time_sleep" "wait_for_gmp_operator" {
-  create_duration = "10s"
+  count           = var.autopilot_cluster ? 1 : 0
+  create_duration = "30s"
 }
 
 # google managed prometheus engine
@@ -23,7 +24,7 @@ resource "helm_release" "gmp-engine" {
   chart            = "${path.module}/charts/gmp-engine/"
   namespace        = var.namespace
   create_namespace = var.create_namespace
-  # timeout increased to support autopilot scaling resources, and give enough time to complete the deployment 
+  # Timeout is increased to guarantee sufficient scale-up time for Autopilot nodes.
   timeout = 1200
   set {
     name  = "projectID"
--- a/applications/rag/main.tf
+++ b/applications/rag/main.tf
@@ -64,20 +64,21 @@ module "infra" {
   source = "../../infrastructure"
   count  = var.create_cluster ? 1 : 0
 
-  project_id        = var.project_id
-  cluster_name      = var.cluster_name
-  cluster_location  = var.cluster_location
-  region            = local.cluster_location_region
-  autopilot_cluster = var.autopilot_cluster
-  private_cluster   = var.private_cluster
-  create_network    = var.create_network
-  network_name      = local.network_name
-  subnetwork_name   = local.network_name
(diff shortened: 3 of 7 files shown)


modules/kuberay-monitoring/main.tf
```diff
@@ -14,7 +14,7 @@
 
 # Temporary workaround to ensure the GMP webhook is installed before applying PodMonitorings.
 resource "time_sleep" "wait_for_gmp_operator" {
-  create_duration = "10s"
+  create_duration = "60s"
```
soorena776, line 17: why not a more gradual increase, like 30s?

artemvmin: I chose 60s because we are in a time crunch. If this solution doesn't work, I would like to know early.

We can always reduce the delay later.

andrewsykim, line 17: Can we gate this behavior to autopilot clusters only?

andrewsykim: (in case we choose to change the default to Standard, we shouldn't add the 60s delay

artemvmin: Keep in mind that the RAG Terraform are deploying a DAG of modules. There are other modules that have longer startups (e.g. CloudSQL instance takes 10+ minutes), and this is happening in parallel. No other modules depend on `kuberay-monitoring.gmp-engine`, and its deployment is quick (~6 seconds), so it will not actually impact the E2E deployment of the RAG app. In other words, as long as this duration is < 10 minutes, there will be no impact to E2E deployment time.

artemvmin: Yeah, that's a good reminder. I meant to do that eventually.

Keep in mind that this 60s does not actually impact E2E time, since CloudSQL is being deployed in parallel and takes 10+ minutes (see comment above).

imreddy13 (approved):

modules/kuberay-monitoring/main.tf
artemvmin: Done.

andrewsykim:
FYI just happened on a recent PR https://github.com/GoogleCloudPlatform/ai-on-gke/pull/468/checks?check_run_id=23176469102

applications/rag/variables.tf
```diff
@@ -31,6 +31,11 @@ variable "cluster_location" {
   type = string
 }
 
+variable "kubernetes_version" {
+  type    = string
+  default = "1.28"
```
andrewsykim, line 36: Does this version still use DPv2? Assuming this was the networking issue we saw, I'm not sure that just downgrading to v1.28 will help

artemvmin: It seems there are multiple features at play. Looks like this satisfies our API requirements and will likely be more stable.

artemvmin:
Standard configuration returned error:

```
Error: Provider produced inconsistent result after apply

When applying changes to
module.kuberay-operator.module.kuberay-workload-identity.google_service_account.cluster_service_account[0],
provider "provider[\"registry.terraform.io/hashicorp/google\"]" produced an
unexpected new value: Root resource was present, but now absent.
```

This appears to be a provider error, likely transient. This PR did not modify the google provider.

PR #467, opened 2024-03-27
Support for Multiple Separate TPU Worker Groups per RayCluster

ryanaoleary:
This PR adds support for multiple worker groups that request TPUs (both single-host and multi-host) in the same Ray Cluster. This PR was tested by manually creating a RayCluster with single-host, multi-host, single-slice, and multi-slice worker groups and verifying that all were injected with the correct `TPU_NAME`, `TPU_WORKER_ID`, `replicaIndex`, and `TPU_WORKER_HOSTNAMES`.

This PR also adds unit tests for `getReplicaIndex` and `getNextWorkerID`, these tests can be run using `go test` from the applications/ray/kuberay-tpu-webhook folder.

### DIFF
--- a/applications/ray/kuberay-tpu-webhook/main.go
+++ b/applications/ray/kuberay-tpu-webhook/main.go
@@ -24,6 +24,7 @@ import (
 type slice struct {
 	clusterName  string
 	groupName    string
+	namespace    string
 	replicaIndex int
 	numOfHosts   int32
 }
@@ -266,13 +267,14 @@ func validateRayCluster(admissionReview *admissionv1.AdmissionReview) (*admissio
 	for i := 0; i < len(workerGroupSpecs); i++ {
 		workerGroupSpec := workerGroupSpecs[i]
 		if containerRequestingTPUs(workerGroupSpec.Template.Spec.Containers...) {
+			klog.V(0).InfoS("validateRayCluster", "RayCluster", namespace+"/"+clusterName, "Worker Group", workerGroupSpec.GroupName, "Requests TPUs", true)
 			// create mapping for pod slices -> TPU_WORKER_HOSTNAMES in cluster
 			replicas := int(*workerGroupSpec.Replicas)
 			numOfHosts := workerGroupSpec.NumOfHosts
 			for replicaIndex := 0; replicaIndex < replicas; replicaIndex++ {
 				// reset past sliceToWorkers and sliceToHostnames entries for slice in ray cluster
 				groupName := workerGroupSpec.GroupName
-				podSlice := slice{clusterName, groupName, replicaIndex, numOfHosts}
+				podSlice := slice{clusterName, groupName, namespace, replicaIndex, numOfHosts}
 				sliceToWorkers[podSlice] = nil
 				sliceToHostnames[podSlice] = ""
 				// generate TPU_WORKER_HOSTNAMES
@@ -284,6 +286,9 @@ func validateRayCluster(admissionReview *admissionv1.AdmissionReview) (*admissio
 					sliceToHostnames[podSlice] = joinedHostNames
 				}
 			}
+		} else {
+			// RayCluster worker group does not request TPUs
+			klog.V(0).InfoS("validateRayCluster", "RayCluster", namespace+"/"+clusterName, "Worker Group", workerGroupSpec.GroupName, "Requests TPUs", false)
 		}
 		// validate NumOfHosts for worker group matches topology nodeSelector
 		workersMatchTopology, err := checkWorkersMatchTopology(clusterName, namespace, workerGroupSpec)
@@ -322,18 +327,28 @@ func getEnvironmentVariable(varName string, container corev1.Container) string {
 	return ""
 }
 
-// get next lowest-index pod slice to assign a pod to in the RayCluster
-// this will be the first pod slice with # created pods < NumOfHosts
+// gets the  next lowest-index pod slice (worker group replica) to assign a pod to in the RayCluster
+// there are three possible cases here:
+//  1. sliceToWorkers is empty, this is the first pod the webhook intercepts
+//     - assign this pod to replica 0
+//  2. The pod slice exists in sliceToWorkers, but has # created workers < NumOfHosts
+//     - assign this pod to the lowest index replica with # created workers < NumOfHosts
+//     - since we update isCreated when a worker is deleted, this allows us to assign re-created
+//     pods to the same replica
+//  3. sliceToWorkers isn't empty, but all slices have # workers == NumOfHosts
+//     - this occurs when the pod we intercept is the first pod of a different slice in the cluster
+//     - we keep track of how many replicas of the same worker group have been added to sliceToWorkers
+//     so far, and assign this pod to the next integer replicaIndex
 func getReplicaIndex(clusterName string, groupName string, namespace string) int {
 	// first pod created in cluster
 	if sliceToWorkers == nil {
 		return 0
 	}
 	nextLowestId := math.MaxInt32
-	numSlices := 0 // tracks # of slices in worker group created so far
+	numReplicas := 0 // tracks # of replicas in worker group created so far
 	for slice, workerList := range sliceToWorkers {
-		if slice.clusterName == clusterName && slice.groupName == groupName {
-			numSlices++
+		if slice.clusterName == clusterName && slice.groupName == groupName && slice.namespace == namespace {
+			numReplicas++
 			createdPods := 0
 			for _, worker := range workerList {
 				if worker.isCreated {
@@ -349,7 +364,7 @@ func getReplicaIndex(clusterName string, groupName string, namespace string) int
 	}
 	// first pod of new slice in cluster
 	if nextLowestId == math.MaxInt32 {
-		nextLowestId = numSlices
+		nextLowestId = numReplicas
 	}
 	klog.V(0).InfoS("getReplicaIndex", "RayCluster", namespace+"/"+clusterName, "Worker Group", groupName, "Replica Index", nextLowestId)
 	return nextLowestId
@@ -443,7 +458,7 @@ func mutatePod(admissionReview *admissionv1.AdmissionReview) (*admissionv1.Admis
 		// assign worker to the next unique ID in the pod slice and update map
 		numOfHosts, _ := getNumTPUHostsFromTopology(clusterName, groupName, namespace, topology, acceleratorType) // ignore error here because topology may not be set yet
 		replicaIndex := getReplicaIndex(clusterName, groupName, namespace)
-		podSlice := slice{clusterName, groupName, replicaIndex, numOfHosts}
+		podSlice := slice{clusterName, groupName, namespace, replicaIndex, numOfHosts}
 		tpuWorkerID := getNextWorkerID(podSlice, namespace, replicaIndex) // defaults to 0 for single-host
 
 		// inject replica index label
@@ -573,7 +588,7 @@ func deletePod(admissionReview *admissionv1.AdmissionReview) (*admissionv1.Admis
 		}
 		// update sliceToWorkers map
 		for slice, _ := range sliceToWorkers {
-			if slice.clusterName == clusterName && slice.groupName == groupName && slice.replicaIndex == replicaIndex {
+			if slice.clusterName == clusterName && slice.groupName == groupName && slice.namespace == namespace && slice.replicaIndex == replicaIndex {
 				// set the pod state to indicate it is not running
 				for index, worker := range sliceToWorkers[slice] {
 					if worker.workerIndex == tpuWorkerID {
--- a/applications/ray/kuberay-tpu-webhook/main.go
+++ b/applications/ray/kuberay-tpu-webhook/main.go
@@ -77,7 +77,7 @@ func containerRequestingTPUs(containers ...corev1.Container) bool {
 	return false
 }
 
-func getNumTPUHostsFromTopology(clusterName string, namespace string, topology string, acceleratorType string) (int32, error) {
+func getNumTPUHostsFromTopology(clusterName string, groupName string, namespace string, topology string, acceleratorType string) (int32, error) {
 	if topology == "" {
 		return 0, errors.New("TPU topology not specified")
 	}
@@ -86,7 +86,7 @@ func getNumTPUHostsFromTopology(clusterName string, namespace string, topology s
 	for i := 0; i < len(topologyVals); i++ {
 		dim, err := strconv.Atoi(topologyVals[i])
 		if err != nil {
-			klog.ErrorS(err, "RayCluster", namespace+"/"+clusterName, "gke-tpu-topology", topology)
+			klog.ErrorS(err, "RayCluster", namespace+"/"+clusterName, "Worker Group", groupName, "gke-tpu-topology", topology)
 			return 0, err
 		}
 		chips *= dim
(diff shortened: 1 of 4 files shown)


applications/ray/kuberay-tpu-webhook/main.go
```diff
@@ -77,7 +77,7 @@ func containerRequestingTPUs(containers ...corev1.Container) bool {
 	return false
 }
```
richardsliu, line 79: Are `clusterName`, `namespace`, and `groupName` only used for logging? It seems a bit odd to me that these are required parameters for a fairly deterministic function.

applications/ray/kuberay-tpu-webhook/main.go
```diff
@@ -263,22 +265,24 @@ func validateRayCluster(admissionReview *admissionv1.AdmissionReview) (*admissio
...
-			groupName := workerGroupSpec.GroupName
-			podSlice := slice{clusterName, groupName, replicaIndex, numOfHosts}
-			sliceToWorkers[podSlice] = nil
-			sliceToHostnames[podSlice] = ""
-			// generate TPU_WORKER_HOSTNAMES
-			if numOfHosts > 1 {
-				joinedHostNames, err := genDNSHostnames(workerGroupSpec, replicaIndex)
-				if err != nil {
-					klog.Error("Failed to generate DNS Hostnames")
+		if containerRequestingTPUs(workerGroupSpec.Template.Spec.Containers...) {
```
richardsliu, line 268: Log the case where the worker group does not request TPUs?

applications/ray/kuberay-tpu-webhook/main.go
```diff
@@ -320,13 +324,16 @@ func getEnvironmentVariable(varName string, container corev1.Container) string {
...
+func getReplicaIndex(clusterName string, groupName string, namespace string) int {
+	// first pod created in cluster
 	if sliceToWorkers == nil {
 		return 0
 	}
 	nextLowestId := math.MaxInt32
+	numSlices := 0 // tracks # of slices in worker group created so far
 	for slice, workerList := range sliceToWorkers {
-		if slice.clusterName == clusterName {
+		if slice.clusterName == clusterName && slice.groupName == groupName {
```
richardsliu, line 335: Should check for namespace as well?

applications/ray/kuberay-tpu-webhook/main.go
```diff
@@ -320,13 +324,16 @@ func getEnvironmentVariable(varName string, container corev1.Container) string {
 
 // get next lowest-index pod slice to assign a pod to in the RayCluster
 // this will be the first pod slice with # created pods < NumOfHosts
-func getReplicaIndex(clusterName string, namespace string) int {
+func getReplicaIndex(clusterName string, groupName string, namespace string) int {
```
richardsliu, line 327: This logic is a bit complex and is critical to the correctness of the webhook. I would suggest:
1. Clearly document the algorithm used here to pick the next replica index;
2. Create a unit test here that captures all of the corner cases we found so far.

applications/ray/kuberay-tpu-webhook/main.go
```diff
@@ -320,13 +324,16 @@ func getEnvironmentVariable(varName string, container corev1.Container) string {
...
+	// first pod created in cluster
 	if sliceToWorkers == nil {
 		return 0
 	}
 	nextLowestId := math.MaxInt32
+	numSlices := 0 // tracks # of slices in worker group created so far
```
richardsliu, line 333: What does "created" mean in this context? The code here increments numSlices regardless of what happens to it.

richardsliu: Suggest using "replica" instead of "slice" to minimize confusion.

applications/ray/kuberay-tpu-webhook/main.go
```diff
@@ -340,10 +347,11 @@ func getReplicaIndex(clusterName string, namespace string) int {
 			}
 		}
 	}
+	// first pod of new slice in cluster
 	if nextLowestId == math.MaxInt32 {
-		klog.ErrorS(errors.New("Replica Index never set"), "RayCluster", namespace+"/"+clusterName, "Replica Index", nextLowestId)
+		nextLowestId = numSlices
```
richardsliu, line 352: When would we get here? If `sliceToWorkers` is populated when we validate the ray cluster, then shouldn't we always find a match?

applications/ray/kuberay-tpu-webhook/main.go
ryanaoleary: Yeah that's the case for `getNumTPUHostsFromTopology`, the logic behind adding them was so that we could have klog Error and Info statements that would be traceable to the right Group, Namespace, and RayCluster if it failed in this function.

applications/ray/kuberay-tpu-webhook/main.go
ryanaoleary: `sliceToWorkers` isn't populated when we validate the RayCluster, `sliceToHostnames` is. We populate `sliceToWorkers` as `TPU_WORKER_ID`s are assigned in `getNextWorkerID`. We get to this if statement when  `sliceToWorkers` isn't empty (so some slices / worker groups have already been added) and we intercept the first pod in a different worker group.

applications/ray/kuberay-tpu-webhook/main.go
ryanaoleary: By "created" we just mean that the slice (i.e. a replica of a TPU worker group) has been added to `sliceToWorkers`, in the case where `NumOfHosts` workers have been added to `sliceToWorkers` for the other replicas of the worker group, `numSlices`  (I'll change it to `numReplicas`) indicates the next `replicaIndex` to assign.

applications/ray/kuberay-tpu-webhook/main.go
ryanaoleary: Done.

applications/ray/kuberay-tpu-webhook/main.go
ryanaoleary: Done.

applications/ray/kuberay-tpu-webhook/main.go
ryanaoleary: I can move the log statements somewhere else, but we should always have `clusterName`, `namespace`, and `groupName` (if not it'd return an error before `getNumTPUHostsFromTopology` so it felt safe to pass them to the function.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+
+func setupTest(t *testing.T) {
+	namespaceStr = "default"
+	instanceName = "raycluster-sample"
+	headNodeIP = "1.2.3.4"
+	groupNameStr = "workergroup"
+	headlessServiceSuffix = "headless-worker-svc"
+
+	// 1 CPU head pod + 1 worker - doesn't request TPUs
+	testCPUPods = []runtime.Object{
```
richardsliu, line 48: Is there a way we could define constant yamls for something like "cpu worker" and "tpu worker", and then re-use these templates in subsequent tests? 

One option is to add a "testdata" directory where you could define these as yaml strings, which you could then parse here to reduce the repetition in test code.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+		},
+	}
+}
+
+func Test_GetReplicaIndex(t *testing.T) {
+	setupTest(t)
+
+	tests := map[string]struct {
+		sliceToWorkers     map[slice][]worker
+		numOfHosts         int32
```
richardsliu, line 460: Maybe call these "expectedNumOfHosts" and "expectedNumReplicas" to be more clear.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+		numReplicas        int
+		additionalGroupStr string
+		numOfHosts2        int32
```
richardsliu, line 463: expectedAdditionalNumOfHosts? to make the association with `additionalGroup` more explicit.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func Test_GetReplicaIndex(t *testing.T) {
```
richardsliu, line 455: One interesting way to test this function is to override the contents of `sliceToWorkers` and seeing how this function behaves in different circumstances.

We can do this in a follow up PR.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+			// for each respective worker group
+			numOfHosts:         4,
+			numReplicas:        4,
+			additionalGroupStr: "another-worker-group",
+			numOfHosts2:        2,
+			numReplicas2:       3,
+		},
+	}
+
+	for name, tc := range tests {
```
richardsliu, line 497: Describe what the test is trying to validate.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+func Test_GetNextWorkerID(t *testing.T) {
+	setupTest(t)
+
+	tests := map[string]struct {
+		numOfHosts  int32
+		numReplicas int
+		deletePodID int
+	}{
+		"single-host, single-slice worker group": {
+			// single-slice, replicaIndex should always be 0
```
richardsliu, line 546: What should the worker ids be here?

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+			// multi-slice, replicaIndex should always be 0-numReplicas for 0-numOfHosts pods
+			numOfHosts:  4,
+			numReplicas: 4,
+			deletePodID: -1,
+		},
+		"deleted pod from multi-host group": {
+			// pod should be reassigned the ID of the deleted pod
+			numOfHosts:  4,
+			numReplicas: 4,
+			deletePodID: 2,
```
richardsliu, line 573: What if you delete multiple pods?

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+		},
+	}
+
+	for name, tc := range tests {
```
richardsliu, line 577: Same thing here, describe what this test validates.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+	testCPUPods                   []runtime.Object
+	testTPUPods                   []runtime.Object
+	testRayClusterAdmissionReview *admissionv1.AdmissionReview
+	testRayClusterNoTPUs          *rayv1.RayCluster
+	testRayClusterSingleHostTPU   *rayv1.RayCluster
+	testRayClusterMultiHostTPU    *rayv1.RayCluster
+	testServices                  []runtime.Object
+	workerSelector                labels.Selector
+	headNodeIP                    string
+	// sliceToWorkers				  map[slice][]worker
```
richardsliu, line 35: Is this intended to use the same copy of `sliceToWorkers` in the webhook?

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+			numReplicas: 1,
+			deletePodID: -1,
+		},
+		"multi-host, multi-slice worker group": {
+			// multi-slice, replicaIndex should always be 0-numReplicas for 0-numOfHosts pods
+			numOfHosts:  4,
+			numReplicas: 4,
+			deletePodID: -1,
+		},
+		"deleted pod from multi-host group": {
```
richardsliu, line 569: Do we need a test for multiple worker groups here?

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
```diff
@@ -0,0 +1,595 @@
...
+			// single-slice, replicaIndex should always be 0
+			numOfHosts:  4,
+			numReplicas: 1,
+		},
+		"multi-host, multi-slice worker group": {
+			// multi-slice, replicaIndex should always be 0-numReplicas for 0-numOfHosts pods
+			numOfHosts:  4,
+			numReplicas: 4,
+		},
+		"multiple worker groups": {
```
richardsliu, line 486: Do we need a test for deleting pods here?

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
ryanaoleary: I think that's a good idea, for now I changed `testCPUPods` and `testTPUPods` to a cpu and tpu worker template whose values we can change within the test. In the next PR with the rest of the tests I think it'd be good to add a testdata directory (since these objects aren't being used yet in this PR).

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
ryanaoleary: These aren't the expected values, these define the test values for `NumOfHosts` and `Replicas` for a hypothetical Ray worker group. The expected values will then be between `0-(NumOfHosts-1)` and `0-(Replicas-1)`.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
ryanaoleary: Done, changed to `additionalNumOfHosts`.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
ryanaoleary: Done.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
ryanaoleary: Done, changed the comment descriptions to define what the worker IDs should be for each test.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
ryanaoleary: Done, added a test case for this.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
ryanaoleary: Done.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
ryanaoleary: Yeah it uses the same `sliceToWorkers` since this is what's added to internally in `getNextWorkerID`. I'll add in some logic to save a deep copy of `sliceToWorkers` and return it to its state after the test. I don't think this should cause any issues unless you run the tests while the webhook is intercepting pods with the same cluster name, namespace, and group name.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
ryanaoleary: Done, added a test case for this to validate that re-created pods will be assigned the correct ID in different worker groups.

applications/ray/kuberay-tpu-webhook/webhook_main_test.go
ryanaoleary: Added a test case to validate that a re-created pod will be assigned to the lowest index replica with created Pods < NumOfHosts.

richardsliu (approved):

PR #476, opened 2024-03-28
Use default Kubernetes version in REGULAR channel

andrewsykim:
In https://github.com/GoogleCloudPlatform/ai-on-gke/pull/466 the Kubernetes version was pinned to v1.28. This PR removes the version override and uses whatever is the default in REGULAR channel


umeshkumhar (approved):

andrewsykim:
Closing since we want to use v1.28 where L4 GPUs are supported by autopilot clusters

PR #481, opened 2024-03-28
Increase ray head node memory and CPU

imreddy13:
- The ray job in the notebook takes 10-20 minutes to run right now because the ray head is doing a lot. Increasing the CPU/mem should help.
- Job runs in ~5 mins now: 
image

### DIFF
--- a/modules/kuberay-cluster/values.yaml
+++ b/modules/kuberay-cluster/values.yaml
@@ -90,16 +90,16 @@ head:
   # for further guidance.
   resources:
     limits:
-      cpu: "4"
+      cpu: "8"
       # To avoid out-of-memory issues, never allocate less than 2G memory for the Ray head.
       # Ray recommends at least 8G memory for production workloads.
-      memory: "8G"
+      memory: "20G"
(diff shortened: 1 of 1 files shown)


andrewsykim:
> The ray job in the notebook takes 10-20 minutes

From my tests so far, it takes about 5m. Which part is taking 10-20m?

andrewsykim:
LGTM to increase size, but I've been consistenly seeing ~5m jobs without the increase. Also is it enough to just increase head node or should we also increase worker node?

andrewsykim:
To be fair, I've been mostly testing with https://github.com/GoogleCloudPlatform/ai-on-gke/pull/482 included in my notebook

modules/kuberay-cluster/values.yaml
```diff
@@ -90,16 +90,16 @@ head:
   # for further guidance.
   resources:
     limits:
-      cpu: "4"
+      cpu: "8"
```
andrewsykim, line 93: The worker node uses similar limits. I wonder if the variation in job time was based on whether the task lands on head node or worker node

imreddy13:
Will revisit this if needed in the future.

PR #482, opened 2024-03-28
Remove uneccesary pip installs from notebook

imreddy13:
Since we now use a custom image that pre installs these packages, we don't need these in the notebook.


andrewsykim (approved):

andrewsykim:
v1.1 cherry-pick https://github.com/GoogleCloudPlatform/ai-on-gke/pull/486

PR #484, opened 2024-03-28
remove csv file write

annapendleton:
custom_messages has bug that causes locust worker restarts. https://github.com/locustio/locust/issues/2608

We rely on custom messages for our custom metrics csv file periodic writes.

Since we've added in exporting the custom metrics to prometheus, I don't think we need the csv file write. The locust runner includes these metrics in its final report. And metrics are viewable via custom dashboards in cloud monitoring.

Rather than trying to make the heartbeat work, I think we should remove the csv file support and migrate to improving the total report that includes locust and GPU metrics as well.

This change also enables custom metrics for all workloads.

### DIFF
--- a/benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py
+++ b/benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py
@@ -56,17 +56,3 @@ class TokenMetricCollector:
         }
         return json.dumps(stats)
 
-    def write_to_csv(self, file_path='custom_metrics.csv'):
-        import csv
-        avg_sent, avg_received, avg_test_time, avg_output_token_latency = self.calculate_average_tokens()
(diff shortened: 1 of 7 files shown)


benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py
kaushikmitr, line 1: maybe in a later PR, we need to flush out this list or it will keep growing indefinitly as they test runs.

kaushikmitr (approved):

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py
annapendleton: +1 that's a great idea - this was originally created with terraform-docs. It would be useful for us to add instructions on how to set that up + how to update the documentation + enforcement on ensuring it was updated.

Issue #425, opened 2024-03-22
Use Ray interactive client in the example notebooks


andrewsykim:
The notebook used for RAG currently writes a python file locally and uses it as an entrypoint to a Ray job using `JobSubmissionClient`. 

Instead of this approach, we can use [Ray's interactive remote client](https://docs.ray.io/en/latest/cluster/running-applications/job-submission/ray-client.html) to run the steps remotely. This would allow the notebook to be broken down into smaller cells that can run in isolation, as opposed to a single job that executes everything at once. 

The trade-off here is that the remote interactivce client may not be as reliable as using a Ray job

Issue #435, opened 2024-03-25
CLOUDSQL_INSTANCE_CONNECTION_NAME env in RayCluster should only be configured for RAG deployments  


andrewsykim:
The CLOUDSQL_INSTANCE_CONNECTION_NAME env var is always set for RayCluster deployments in `applications/ray`. However, this env var is only needed for the RAG deployment. This env var should only be set when `applications/ray` is used by the RAG deployment.

PR #504, opened 2024-03-29
Refactor Rag notebook

richardsliu:
Refactored the RAG notebook to be more modular and documented. 

* Moved the CloudSQL code into the notebook instead of running on a Ray Clusters
* Execute remote Ray tasks instead of using a job submit with busy waiting
* Modified the CloudSQL code to use bulk insert

Fixes https://github.com/GoogleCloudPlatform/ai-on-gke/issues/425

Preview: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/rag-notebook/applications/rag/example_notebooks/rag-kaggle-ray-sql-refactored.ipynb


modules/jupyter/jupyter_config/config-selfauth-autopilot.yaml
```diff
@@ -88,6 +88,7 @@ singleuser:
   extraEnv:
     # Used for GCSFuse to set the ephemeral storage as the home directory. If not set, it will show a permission error on the pod log when using GCSFuse.
     JUPYTER_ALLOW_INSECURE_WRITES: "true"
+    CLOUDSQL_INSTANCE_CONNECTION_NAME: ${cloudsql_instance_connection_name}
```
andrewsykim, line 91: btw. this is going to add the CLOUDSQL env var and secrets for the Jupyter-only installation (not RAG)

andrewsykim: (likewise for the secret mounts)

applications/rag/example_notebooks/rag-kaggle-ray-sql-refactored.ipynb
```diff
@@ -0,0 +1,449 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a814e91b-3afe-4c28-a3d6-fe087c7af552",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!pip install ray[default]==2.9.3 kaggle==1.6.6\n",
+    "!pip install langchain==0.1.10 ray==2.9.3 datasets sentence-transformers\n",
```
andrewsykim, line 51: I think there were some issues in the past with some of these packages taking way too long to install. Specifically sentence-transformers. Are we going to bake the dependencies into the jupyter image?

applications/rag/example_notebooks/rag-kaggle-ray-sql-refactored.ipynb
```diff
@@ -0,0 +1,449 @@
+{
```
andrewsykim, line 1: Can you name this file `rag-kaggle-ray-sql-interactive.ipynb` instead?

modules/jupyter/jupyter_config/config-selfauth-autopilot.yaml
andrewsykim: ^ we spoke offline -- I meant that jupyer only installs will set these environment varibles

andrewsykim (commented):
Overall LGTM, this is a great improvement!

I think the following issues need to be addressed if we're going to cherry-pick this into release-1.1:
* Dependency install time and whether a custom image is warranted. 
* Leaking CloudSQL env vars and secret mounts into all Jupyter installments -- these should only be set when using Jupyter for RAG.
* If we choose to keep both versions of the notebook, please add the markdown descriptions in the other notebook as well

modules/jupyter/jupyter_config/config-selfauth-autopilot.yaml
richardsliu: Made these variables default to empty string. If not explicitly passed, the env var will be empty.

applications/rag/example_notebooks/rag-kaggle-ray-sql-refactored.ipynb
richardsliu: I've tested and it turns out the notebook does need to install langchain and sentence-transformers. I can see if we could use a custom image here to skip these pip installs.

richardsliu:
> * Dependency install time and whether a custom image is warranted.

The notebook needs to install `sentence-transformer` and `langchain`. I'll see if we can use a custom image to reduce the install time.

> * Leaking CloudSQL env vars and secret mounts into all Jupyter installments -- these should only be set when using Jupyter for RAG (likewise for Ray ref: [CLOUDSQL_INSTANCE_CONNECTION_NAME env in RayCluster should only be configured for RAG deployments   #435](https://github.com/GoogleCloudPlatform/ai-on-gke/issues/435)).

This is now fixed.

> * If we choose to keep both versions of the notebook, please add the markdown descriptions in the other notebook as well

I prefer to keep this PR limited to changes relevant for the new notebook. Adding markdown to the older notebook wouldn't be too interesting since all the code is executed in one cell.

applications/rag/example_notebooks/rag-kaggle-ray-sql-refactored.ipynb
richardsliu: Changed to a custom image with dependencies baked in. Removed this section from the notebook.

applications/rag/example_notebooks/rag-kaggle-ray-sql-refactored.ipynb
```diff
@@ -0,0 +1,449 @@
...
+   "id": "ba9551ec-883e-4bde-9f12-f663aedc12e5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import ray\n",
+    "\n",
+    "ray.init(\n",
+    "    address=\"ray://ray-cluster-kuberay-head-svc:10001\",\n",
+    "    runtime_env={\n",
+    "        \"pip\": [               \n",
```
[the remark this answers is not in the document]
richardsliu: I think Ray will skip pip installing these if the image is already on the node. When I tested this, this step usually finishes in a few seconds.

applications/rag/example_notebooks/rag-kaggle-ray-sql-interactive.ipynb
```diff
@@ -0,0 +1,425 @@
...
+   "metadata": {},
+   "source": [
+    "# RAG-on-GKE Application\n",
+    "\n",
+    "This is a Python notebook for generating the vector embeddings used by the RAG on GKE application. For full information, please checkout the GitHub documentation [here](https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/rag/README.md).\n",
+    "\n",
+    "\n",
+    "## Setup Kaggle Credentials\n",
+    "\n",
+    "First we will setup your Kaggle credentials. Replace the following with your own settings from the Kaggle web page. Navigate to https://www.kaggle.com/settings/account and generate an API token to be used to setup the env variable. See https://www.kaggle.com/docs/api#authentication how to create one."
```
imreddy13, line 15: nit: "to be used to setup the env variable" is kind of weird. Maybe just have the instructions for how to create Kaggle creds and ask users to replace the specific variables (KAGGLE_USERNAME and KAGGLE_KEY) in the cell below with their Kaggle creds?

modules/jupyter/jupyter_config/config-selfauth-autopilot.yaml
```diff
@@ -111,9 +112,17 @@ singleuser:
...
             mountOptions: "implicit-dirs,uid=1000,gid=100"
+      - name: secret-volume
+        secret:
+          secretName: ${secret_name}
+          optional: true
+
     extraVolumeMounts:
         - name: data-vol
-          mountPath: /persist-data
+          mountPath: /data
```
imreddy13, line 122: we don't need persist-data anymore? awesome! this always confused me.

modules/jupyter/main.tf
```diff
@@ -17,6 +17,7 @@ data "google_project" "project" {
 }
 
 locals {
+  cloudsql_instance_connection_name = var.cloudsql_instance_name != "" ? format("%s:%s:%s", var.project_id, var.db_region, var.cloudsql_instance_name) : ""
```
imreddy13, line 20: could you please try to run tf apply from /applications/jupyter to ensure that the secrets mounting isn't breaking just the jupyter installation? it should be fairly quick to test

modules/jupyter/variables.tf
```diff
@@ -17,6 +17,18 @@ variable "namespace" {
   description = "Kubernetes namespace where resources are deployed"
 }
 
+variable "notebook_image" {
+  type        = string
+  description = "Jupyter notebook image name"
+  default     = "jupyter/tensorflow-notebook"
```
imreddy13, line 23: In addition to defaulting, can you also explicitly set this value from /applications/jupyter/main.tf so it's clear which app is using which image?

modules/jupyter/variables.tf
```diff
@@ -17,6 +17,18 @@ variable "namespace" {
...
+}
+
+variable "notebook_image_tag" {
+  type        = string
+  description = "Jupyter notebook image tag"
+  default     = "python-3.10"
```
imreddy13, line 29: same

applications/rag/example_notebooks/rag-kaggle-ray-sql-interactive.ipynb
richardsliu: Merged the cells.

modules/jupyter/jupyter_config/config-selfauth-autopilot.yaml
richardsliu: Glad you agree.

modules/jupyter/variables.tf
richardsliu: Done.

modules/jupyter/main.tf
richardsliu: Good call. Fixed an error.

andrewsykim:
> I prefer to keep this PR limited to changes relevant for the new notebook. Adding markdown to the older notebook wouldn't be too interesting since all the code is executed in one cell.

Should we simplify and remove the old notebook or keep both? With the changes in this PR, do both notebooks work?

richardsliu:
> > I prefer to keep this PR limited to changes relevant for the new notebook. Adding markdown to the older notebook wouldn't be too interesting since all the code is executed in one cell.
> 
> Should we simplify and remove the old notebook or keep both? With the changes in this PR, do both notebooks work?

I fixed some minor issues in the old notebook and added a header. We should probably keep it as a backup.

imreddy13 (approved):
Thank you!

PR #512, opened 2024-04-01
Add model, inference server and source identifiers

achandrasekar:
This change adds labels to identify the specific model, inference server and the source of the yaml file.

### DIFF
--- a/benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
+++ b/benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
@@ -14,6 +14,7 @@ spec:
     metadata:
       labels:
         app: locust-worker
+        examples.ai.gke.io/source: ai-on-gke-benchmarks
     spec:
       serviceAccountName: ${ksa}
       containers:
--- a/benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-master-controller.yaml.tpl
+++ b/benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-master-controller.yaml.tpl
@@ -14,6 +14,7 @@ spec:
     metadata:
       labels:
         app: locust-master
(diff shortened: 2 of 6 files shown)


benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
```diff
@@ -14,6 +14,7 @@ spec:
     metadata:
       labels:
         app: locust-worker
+        examples.ai.gke.io/source: ai-on-gke-benchmarks
```
annapendleton, line 17: Couple things I'm interested in adding here if possible:

"$tokenizer" var would give us the ai.gke.io/model here if we also wanted to know what models are being benchmarked

likewise, "$inference_server_framework" var would give us ai.gke.io/inference-server if we wanted to know what inference backends were being used for the benchmarking tooling.

annapendleton (approved):

benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
achandrasekar: For models, we want to count only the model server instances for CPU and GPU usage. Benchmarking tool is not essentially running the model but sending requests to it. So, we are excluding it not to overcount.

annapendleton: My understanding is that customers may use the benchmarking tool without using the automatically deployed model servers. Eg. in the byo-model server, we'd also get information on what model servers they chose to deploy themselves and what models they're benchmarking in this scenario. But possibly a niche enough case that we don't care to cover in our usage metrics (eg. may not provide enough value to be worth adding).

Double counting is defs an issue unless we make the labels more granular for each section of the benchmarking suite (eg. infra, inference-server, benchmarking)

PR #511, opened 2024-03-31
Add tests for RAG prompt

andrewsykim:

PR #513, opened 2024-04-01
Update frontend image

blackzlq:
1. Remove enabled filter for two filters.
2. Add more description on the filters

Tested-by: zlq

### DIFF
--- a/applications/rag/frontend/container/static/script.js
+++ b/applications/rag/frontend/container/static/script.js
@@ -48,27 +48,19 @@ function onReady() {
         enableForm(false);
 
         // Collect filter data
-        var filterEnabled = document.getElementById("toggle-filters").checked;
-        var body = JSON.stringify({
+        let data = {
             prompt: prompt,
-        })
-        if (filterEnabled) {
-            let data = {
-                prompt: prompt,
-            }
-
-            if (document.getElementById('toggle-nlp-filter-section').checked) {
-                data.nlpFilterLevel = document.getElementById("nlp-range").value
-            }
-
-            if (document.getElementById('toggle-dlp-filter-section').checked) {
-                data.inspectTemplate = document.getElementById('inspect-template-dropdown').value;
-                data.deidentifyTemplate = document.getElementById('deidentify-template-dropdown').value;
-            }
-
-            body = JSON.stringify(data)
         }
 
+        if (document.getElementById('toggle-nlp-filter-section').checked) {
+            data.nlpFilterLevel = document.getElementById("nlp-range").value
+        }
+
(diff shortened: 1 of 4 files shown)


applications/rag/frontend/container/templates/index.html
```diff
@@ -35,36 +35,40 @@
...
+        </select>
+        <p id="deidentify-template-msg" style="display: none;">Please go to your project to create the de-identification template.</p>
+      </div>
     </div>
 
     <!-- Checkbox to toggle nlp-filter-section -->
-    <div id="toggle-nlp-filter-section-div" style="display: none;">
-      <label for="toggle-nlp-filter-section">Show NLP Filter Range:</label>
+    <div id="toggle-nlp-filter-section-div">
+      <label for="toggle-nlp-filter-section"><a href="https://cloud.google.com/natural-language/docs/moderating-text">Text moderation</a>  analyzes a document against a list of safety attributes, which include "harmful categories" and topics that may be considered sensitive.:</label>
```
imreddy13, line 65: Can we delete the colon at the end?

imreddy13 (approved):
thank you!

applications/rag/frontend/container/static/script.js
```diff
@@ -48,27 +48,19 @@ function onReady() {
         enableForm(false);
 
         // Collect filter data
-        var filterEnabled = document.getElementById("toggle-filters").checked;
-        var body = JSON.stringify({
+        let data = {
             prompt: prompt,
```
andrewsykim, line 52: @blackzlq I'm adding some prompts to test in here https://github.com/GoogleCloudPlatform/ai-on-gke/pull/511 

Can you suggest some ways to test prompts with DLP/NLP?

blackzlq: Yeah, @imreddy13 sends the doc in chat room.

applications/rag/frontend/container/templates/index.html
blackzlq: Good catch!

applications/rag/frontend/container/static/script.js
blackzlq: who worked with Robert De Niro and name one film they collaborated? with name filter to select the info_type personal_name and de-identity template to replace the content with info_type name. is good for DLP

and any prompt then set the range bar to 100 should return `The responses is deemed inapproriate for display`

`Which movie will show blowing up a building ` with range value 60 should got the `The responses is deemed inapproriate for display` and if we lower the value it should return the original response.

blackzlq:
There is no way for this change affect existing test, we haven't update the image SHA, rerun the test

Issue #502, opened 2024-03-28
Remove 60s sleep waiting for GMP


andrewsykim:
Follow-up to https://github.com/GoogleCloudPlatform/ai-on-gke/pull/497/files#r1543631746

german-grandas:
Removed on https://github.com/GoogleCloudPlatform/ai-on-gke/pull/781

PR #514, opened 2024-04-01
Update main.tf

artemvmin:
The `count` parameters is calculated when generating helm resource dependency. Therefore, it cannot depend on values that are unknown at apply time (e.g. on a data source field for a cluster which is not yet created).

### DIFF
--- a/modules/kuberay-monitoring/main.tf
+++ b/modules/kuberay-monitoring/main.tf
@@ -14,8 +14,6 @@
 
 # Temporary workaround to ensure the GMP webhook is installed before applying PodMonitorings.
 resource "time_sleep" "wait_for_gmp_operator" {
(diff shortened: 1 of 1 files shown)


modules/kuberay-monitoring/main.tf
```diff
@@ -14,8 +14,6 @@
 
 # Temporary workaround to ensure the GMP webhook is installed before applying PodMonitorings.
 resource "time_sleep" "wait_for_gmp_operator" {
-  ## Temporary workaroud, This is impacting for the standard cluster dependency graph
-  #count           = var.autopilot_cluster ? 1 : 0
```
andrewsykim, line 18: FYI: https://github.com/GoogleCloudPlatform/ai-on-gke/issues/502

PR #515, opened 2024-04-01
Update domain to support wildcard DNS

umeshkumhar:
- Remove all references of nip.io
- add support as wildcardDNS, that means for better development experience, dev can provide domain inputs as ***.nip.io** that will spin up application specific static-ip and setup nip.io based sub-domains.

### DIFF
--- a/applications/rag/frontend/variables.tf
+++ b/applications/rag/frontend/variables.tf
@@ -129,8 +129,8 @@ variable "support_email" {
 
 variable "domain" {
   type        = string
-  description = "Provide domain for ingress resource and ssl certificate. If it's empty, it will use nip.io wildcard dns"
-  default     = ""
+  description = "Provide domain for ingress resource and ssl certificate."
+  default     = "frontend.example.com"
 }
 
 variable "client_id" {
--- a/applications/rag/metadata.yaml
+++ b/applications/rag/metadata.yaml
@@ -97,7 +97,7 @@ spec:
       - name: frontend_domain
         description: Domain used for application and SSL certificate. 
         varType: string
-        defaultValue: "rag.example.com"
+        defaultValue: "frontend.example.com"
       - name: frontend_k8s_backend_config_name
         description: Name of the Kubernetes Backend Config
         varType: string
--- a/applications/rag/variables.tf
+++ b/applications/rag/variables.tf
@@ -152,8 +152,8 @@ variable "frontend_k8s_backend_service_port" {
 
 variable "frontend_domain" {
   type        = string
-  description = "Domain used for SSL certificate. If it's empty, *.nip.io DNS is used."
-  default     = ""
+  description = "Domain used for SSL certificate."
+  default     = "frontend.example.com"
 }
 
 variable "frontend_client_id" {
@@ -211,8 +211,8 @@ variable "ray_dashboard_k8s_backend_service_port" {
 
 variable "ray_dashboard_domain" {
   type        = string
-  description = "Domain used for SSL certificate. If it's empty, *.nip.io DNS is used."
-  default     = ""
+  description = "Domain used for SSL certificate."
+  default     = "ray.example.com"
 }
 
 variable "ray_dashboard_client_id" {
@@ -278,8 +278,8 @@ variable "jupyter_k8s_backend_service_port" {
 
 variable "jupyter_domain" {
   type        = string
-  description = "Domain used for SSL certificate. If it's empty, *.nip.io DNS is used."
-  default     = ""
+  description = "Domain used for SSL certificate."
+  default     = "jupyter.example.com"
 }
 
 variable "support_email" {
--- a/applications/rag/workloads.tfvars
+++ b/applications/rag/workloads.tfvars
@@ -58,15 +58,15 @@ dataset_embeddings_table_name = "netflix_reviews_db"
 
 ## Jupyter IAP Settings
 jupyter_add_auth          = false                                                                 # Set to true to enable authenticated access via IAP.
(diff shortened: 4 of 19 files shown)


applications/rag/frontend/variables.tf
```diff
@@ -129,8 +129,8 @@ variable "support_email" {
 
 variable "domain" {
   type        = string
-  description = "Provide domain for ingress resource and ssl certificate. If it's empty, it will use nip.io wildcard dns"
-  default     = ""
+  description = "Provide domain for ingress resource and ssl certificate."
+  default     = "frontend.example.com"
```
bjornsen, line 133: I don't think we should default this domain to `example.com`. We should default to blank. It's fine to use in the example `.tfvars` since those are user facing and meant to be edited. I also suspect that defaulting to this might break things or have unexpected issues down the line.

applications/rag/variables.tf
```diff
@@ -152,8 +152,8 @@ variable "frontend_k8s_backend_service_port" {
 
 variable "frontend_domain" {
   type        = string
-  description = "Domain used for SSL certificate. If it's empty, *.nip.io DNS is used."
-  default     = ""
+  description = "Domain used for SSL certificate."
+  default     = "frontend.example.com"
```
bjornsen, line 156: Same comments on above of defaulting this to `""` and not using `example.com` as a default. Also, I'd recommending of having multiple levels of the same defaults because it makes changes like this more complicated in the future.

applications/rag/frontend/variables.tf
umeshkumhar: Okay we can have that.. This particluar is needed for marketplace defaults..

applications/rag/variables.tf
umeshkumhar: Okay we can have that.. This particluar is needed for marketplace defaults..

applications/rag/frontend/variables.tf
imreddy13: Why does "frontend.example.com" interrupt the terraform execution, but not ""? Wouldn't the managed cert provisioning fail for ""?

bjornsen: Can we just turn off IAP with an empty domain? Using `frontend.example.com` isn't going to give you a functioning endpoint.

imreddy13: Oh, I misread what Umesh wrote. Yes, I agree that we should not allow the execution to succeed for a fake domain, that's a worse experience than failing early.

Might not be trivial to turn IAP off, but I think the managed cert creation will fail anyway with a specific error if you don't provide a valid domain string. So I'm ok with "".

umeshkumhar: okay lets keep it empty.. and on `""` domain will skip managedcertificate creation.

umeshkumhar: actually, let me include variable validations,, that means if domain is kept empty or has example.com.. then fail the terraform at plan itself.. and fail fast.. so user knows that if IAP is enabled.. then domain is required.

applications/rag/workloads.tfvars
```diff
@@ -58,15 +58,15 @@ dataset_embeddings_table_name = "netflix_reviews_db"
...
 jupyter_add_auth          = false                                                                 # Set to true to enable authenticated access via IAP.
-jupyter_domain            = ""                                                                    # Custom domain for ingress resource and ssl certificate. If empty, it will use nip.io wildcard DNS.
+jupyter_domain            = "jupyter.example.com"                                                 # Custom domain for ingress resource and ssl certificate. 
 jupyter_members_allowlist = "user:<email>,group:<email>,serviceAccount:<email>,domain:google.com" # Allowlist principals for access.
 
 ## Frontend IAP Settings
 frontend_add_auth          = false                                                                 # Set to true to enable authenticated access via IAP.
-frontend_domain            = ""                                                                    # Custom domain for ingress resource and ssl certificate. If empty, it will use nip.io wildcard DNS.
+frontend_domain            = "frontend.example.com"                                                # Custom domain for ingress resource and ssl certificate.
```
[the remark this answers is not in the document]
umeshkumhar: keeping as reference example.

andrewsykim:
From CI:

```
When applying changes to
module.jupyterhub.module.jupyterhub-workload-identity.google_service_account.cluster_service_account[0],
provider "provider[\"registry.terraform.io/hashicorp/google\"]" produced an
unexpected new value: Root resource was present, but now absent.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.

Warning: "default_secret_name" is no longer applicable for Kubernetes v1.24.0 and above

  with module.frontend.module.frontend-workload-identity.kubernetes_service_account.main[0],
  on .terraform/modules/frontend.frontend-workload-identity/modules/workload-identity/main.tf line 51, in resource "kubernetes_service_account" "main":
  51: resource "kubernetes_service_account" "main" {

Starting from version 1.24.0 Kubernetes does not automatically generate a
token for service accounts, in this case, "default_secret_name" will be empty

(and one more similar warning elsewhere)
```

Can't tell if it's introduced from changes in this PR

imreddy13 (approved):

applications/rag/metadata.yaml
```diff
@@ -97,7 +97,7 @@ spec:
       - name: frontend_domain
         description: Domain used for application and SSL certificate. 
         varType: string
-        defaultValue: "rag.example.com"
+        defaultValue: "frontend.example.com"
```
bjornsen, line 100: Is this the default value for the terraform templates? May need to be updated to "".

bjornsen (approved):

PR #487, opened 2024-03-28
Improve logic for polling Ray job status in jupyter notebook

andrewsykim:
Before:
```
Job submitted with ID: raysubmit_mSZ1a5U3QTL39dXN
Job status: PENDING
Job info: Job has not started yet. It may be waiting for the runtime environment to be set up.
Job status: RUNNING
Job info: Job is currently running.
<wait 5 mins>
Job  completed in 600 seconds.
```

After:
```
Job submitted with ID: raysubmit_mSZ1a5U3QTL39dXN
Job status: PENDING
Job info: Job has not started yet. It may be waiting for the runtime environment to be set up.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
Job info: Job is currently running.
Job status: RUNNING
[...]

PR #517, opened 2024-04-01
Revert "Improve logic for polling Ray job status in jupyter notebook"

artemvmin:
Reverts GoogleCloudPlatform/ai-on-gke#487, as it is adding unnecessary spam to the command output.


andrewsykim:
Likely obsolete depending on whether https://github.com/GoogleCloudPlatform/ai-on-gke/pull/504 merges

PR #525, opened 2024-04-02
Use different SA names for Ray, Jupyter and RAG

imreddy13:
This allows creating Ray and Jupyter applications in the same project, and also doesn't accidentally delete SAs for the wrong deployment during tf destroy.

### DIFF
--- a/applications/rag/variables.tf
+++ b/applications/rag/variables.tf
@@ -52,7 +52,7 @@ variable "additional_labels" {
 variable "jupyter_service_account" {
   type        = string
   description = "Google Cloud IAM service account for authenticating with GCP services"
-  default     = "jupyter-sa"
+  default     = "jupyter-rag-sa"
 }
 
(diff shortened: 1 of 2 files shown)


applications/jupyter/variables.tf
```diff
@@ -46,7 +46,7 @@ variable "gcs_bucket" {
 variable "workload_identity_service_account" {
   type        = string
   description = "workload identity service account"
-  default     = "jupyter-sa"
+  default     = "jupyter-app-sa"
```
andrewsykim, line 49: It probably makes more sense to keep this `jupyter-sa` and update the RAG implementation to use `jupyter-rag-sa` and `ray-rag-sa`. Thoughts?

imreddy13: Done!

andrewsykim (approved):

PR #534, opened 2024-04-02
Alter locust to not require a hf token

rsgowman:
Additionally, changed how it does use one to be identical to tgi, including using secret manager (rather than a k8s secret, which would end up being stored in tf state.)

Also bump version of transformers to support google/gemma-2b

### DIFF
--- a/benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
+++ b/benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
@@ -44,8 +44,10 @@ spec:
               value: ${tokenizer}
             - name: USE_BEAM_SEARCH
               value: ${use_beam_search}
+%{ for hugging_face_token_secret in hugging_face_token_secret_list ~}
             - name: HUGGINGFACE_TOKEN
               valueFrom:
                 secretKeyRef:
-                  name: ${huggingface_secret}  # Replace ${huggingface_secret} with your secret's name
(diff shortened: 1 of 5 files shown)


benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
```diff
@@ -44,8 +44,10 @@ spec:
               value: ${tokenizer}
             - name: USE_BEAM_SEARCH
```
annapendleton, line 45: Oh interesting, I'm surprised we don't need the secret volume mount like we do in TGI: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/7c5ca2808c502c7d63cc0d7ebe05957f95e967c2/benchmarks/inference-server/text-generation-inference/manifest-templates/text-generation-inference.tftpl#L54

do you know why that is? Maybe in a separate PR we can clean up the TGI implementation if it's not required

annapendleton (approved):

benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
rsgowman: AFAICT, tgi supplies the secret twice (once in an env var, again in /var/secrets) though not being familiar enough with tgi, I don't know why. My guess is that the volume mount is not used, but we'd need to try it out.

PR #536, opened 2024-04-02
Bump GMP delay to 60 seconds

artemvmin:
The bump to 30 seconds reduced flakiness to nearly zero cases over the last 5 days. Doubling to 60 seconds to ensure we don't hit this during demo season. We already have a long-term fix underway to replace this.

### DIFF
--- a/modules/kuberay-monitoring/main.tf
+++ b/modules/kuberay-monitoring/main.tf
@@ -14,7 +14,7 @@
 
 # Temporary workaround to ensure the GMP webhook is installed before applying PodMonitorings.
 resource "time_sleep" "wait_for_gmp_operator" {
-  create_duration = "30s"
+  create_duration = "60s"
 }
 
 # google managed prometheus engine


imreddy13 (approved):

modules/kuberay-monitoring/main.tf
```diff
@@ -14,7 +14,7 @@
 
 # Temporary workaround to ensure the GMP webhook is installed before applying PodMonitorings.
 resource "time_sleep" "wait_for_gmp_operator" {
-  create_duration = "30s"
+  create_duration = "60s"
```
andrewsykim, line 17: I'm not 100% convinced that increasing this timeout actually helps. When we increased from 10s to 30s I didn't see flakes reduce. I'm more hesitant about it because it now also runs in Standard and not exclusive to Autopilot. 

Can we instead look into changing the dependency graph so that we attempt the monitoring installation sometime later when we know GMP is running?

artemvmin: > When we increased from 10s to 30s I didn't see flakes reduce.

We did, drastically. I am getting my data from cloud build runs, so the sample size is pretty good. As I mentioned, we only got 1 gmp-related error in the last 5 days, whereas we were seeing it every couple of runs prior to that.

imreddy13: @andrewsykim We spot checked the last few runs and I don't see the GMP issue anymore - so it seems like the timeout did help. @yiyinglovecoding is working on adding scheduled nightly runs to the main and release-1.1 branches which should give us more reliable data than people's PRs.

PR #542, opened 2024-04-03
Add Validation for IAP domain and allowlist

umeshkumhar:
- Add checks that will fail fast, if domain and member_allowlist are empty..

### DIFF
--- a/modules/iap/iap.tf
+++ b/modules/iap/iap.tf
@@ -33,7 +33,7 @@ data "null_data_source" "domain_validation" {
 
     precondition {
       condition     = length(var.members_allowlist) != 0
-      error_message = "IAP configuration requires valid members_allowlist, Please provide a valid input for ${var.app_name} application."
+      error_message = "IAP configuration requires allowlisting users. Please provide a valid allowlist for ${var.app_name} application."
     }
   }
 }
--- a/modules/iap/iap.tf
+++ b/modules/iap/iap.tf
@@ -28,12 +28,12 @@ data "null_data_source" "domain_validation" {
   lifecycle {
     precondition {
       condition     = var.domain != ""
-      error_message = "IAP configured requires domain name, Please provide a valid domain name for ${var.app_name} application."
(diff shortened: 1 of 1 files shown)


modules/iap/iap.tf
```diff
@@ -21,6 +21,23 @@ data "google_project" "project" {
...
 }
 
+## Check if domain and members_allowlist is provided
+data "null_data_source" "domain_validation" {
+  inputs = {}
+
+  lifecycle {
+    precondition {
+      condition     = var.domain != ""
+      error_message = "IAP configured requires domain name, Please provide a valid domain name for ${var.app_name} application."
```
imreddy13, line 31: "IAP configuration"

modules/iap/iap.tf
```diff
@@ -21,6 +21,23 @@ data "google_project" "project" {
...
+    }
+
+    precondition {
+      condition     = length(var.members_allowlist) != 0
+      error_message = "IAP configured requires valid members_allowlist, Please provide a valid input for ${var.app_name} application."
```
imreddy13, line 36: "IAP configuration requires allowlisting users. Please provide a valid allowlist for ${var.app_name} application."

imreddy13 (approved):

modules/iap/iap.tf
```diff
@@ -21,6 +21,23 @@ data "google_project" "project" {
...
+  inputs = {}
+
+  lifecycle {
+    precondition {
+      condition     = var.domain != ""
+      error_message = "IAP configuration requires domain name, Please provide a valid domain name for ${var.app_name} application."
+    }
+
+    precondition {
+      condition     = length(var.members_allowlist) != 0
```
blackzlq, line 35: member allowlist can be empty, they have the right to add them later

imreddy13: I think this just makes the experience more confusing, most users would want to allowlist at least themselves - they'll leave this blank and wonder why they can't login later. Prefer to fail early and force them to add something to the allowlist.

andrewsykim:
LGTM

PR #544, opened 2024-04-03
Add vector database section and demo 

bkauf:
Add vector database section and demo


jsoohoo-google (approved):
Approved - new tutorial that does not affect modules/apps folders. All presubmit tests pass.

PR #549, opened 2024-04-03
add deployment name prefix for cluster_name in MP

umeshkumhar:
Add deployment name as prefix to clustername in marketplace deployment

### DIFF
--- a/applications/rag/metadata.yaml
+++ b/applications/rag/metadata.yaml
@@ -46,7 +46,7 @@ spec:
         defaultValue: ""
       - name: cluster_name
         varType: string
-        required: true
+        defaultValue: "ai-on-gke"
       - name: create_brand
         description: Create Brand OAuth Screen
         varType: bool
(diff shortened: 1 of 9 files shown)


applications/rag/metadata.yaml
```diff
@@ -46,7 +46,7 @@ spec:
         defaultValue: ""
       - name: cluster_name
         varType: string
-        required: true
+        defaultValue: "ai-on-gke"
```
andrewsykim, line 49: While you're at it, can you also update kubernetes_namespace to default to `ai-on-gke` to match the default value used for Ray and Jupyter

applications/ray/metadata.yaml
```diff
@@ -39,7 +39,7 @@ spec:
         defaultValue: ""
       - name: cluster_name
         varType: string
-        required: true
```
andrewsykim, line 42: (not blocking)

Add a description here:
```
The name of the GKE cluster. The deployment solution name will be prefixed in the final cluster name.
```

Or something like that

PR #557, opened 2024-04-04
Revert "Add Validation for IAP domain and allowlist"

imreddy13:
Reverts GoogleCloudPlatform/ai-on-gke#542

Allowlist check is broken


andrewsykim:
LGTM

PR #563, opened 2024-04-04
tpu-provisioner: Syncing JobSet Pod startup via a ConfigMap volume

nstogner:
Facilitate semi-synchronous startups for Pods in JobSets. A ConfigMap is used that communicates when all Pods have been started.


tpu-provisioner/internal/controller/pods_started_controller.go
```diff
@@ -0,0 +1,171 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// PodsStartedReconciler watches Pods and updates a ConfigMap when a minimum number of
```
nstogner, line 36: @danielvegamyhre

tpu-provisioner/internal/controller/pods_started_controller.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	jobsetRestart, ok := labels[jobsetRestartLabel]
+	if !ok {
+		return ctrl.Result{}, nil
+	}
+
+	// Assume the ConfigMap is present (created alongside the JobSet).
+	// If the ConfigMap is set as an optional volume in the Pod spec
+	// the propagation of a ConfigMap creation into container mount
+	// appears to take a long time (>30s in a test).
```
danielvegamyhre, line 108: I like this approach! 

As an aside, apparently the container mount point update takes up to a full kubelet sync (1min) so that is the cause of the delay here I believe: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically

We can potentially force it to mount faster by updating a pod annotation

tpu-provisioner/internal/controller/pods_started_controller.go
```diff
@@ -0,0 +1,171 @@
...
+	client.Client
+	Scheme   *runtime.Scheme
+	Recorder record.EventRecorder
+
+	Provider cloud.Provider
+}
+
+//+kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+
+func (r *PodsStartedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
danielvegamyhre, line 52: I think maybe this could be optimized by only reconciling on leader pods (completion index 0), then do the same "threshold check" logic, and if we fail this threshold check, re-enqueue the pod for reconciliation again after some short interval. What do you think?

tpu-provisioner/internal/controller/pods_started_controller.go
```diff
@@ -0,0 +1,171 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+// PodsStartedReconciler watches Pods and updates a ConfigMap when a minimum number of
+// associated Pods have started.
+// It assumes that these Pods are managed by a JobSet, otherwise it will ignore them.
+// The threshold is set via an annotation and is intended to be the total number of Pods
+// in the JobSet - allowing for semi-synchronous startup across an entire JobSet.
+// This threshold annotation is used to avoid a lookup of the number of replicas in the JobSet.
```
danielvegamyhre, line 41: If we only reconcile on leader pods (see comment below), I think it would be fine to look up the JobSet and derive the expected pod count from the spec, rather than hard coding it in the annotation. Maybe we can user a Lister to maintain a local cache with the JobSet object?

This would improve the user-experience IMO, we would need to validate it works at scale though.

tpu-provisioner/internal/controller/pods_started_controller.go
nstogner: Actually thinking I will combine both this suggestion and the other one to look up the JobSet. If I watch Pods but reconcile on JobSets then I can accomplish both: 1. Dedup'ing and rate limiting reconciles for a given JobSet, and 2. Having natural access to JobSet information.

tpu-provisioner/internal/controller/synced_start_controller.go
```diff
@@ -0,0 +1,223 @@
...
+				return ctrl.Result{}, fmt.Errorf("setting owner reference: %w", err)
+			}
+			if err := r.Create(ctx, &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      configMapName,
+					Namespace: req.Namespace,
+				},
+			}); err != nil {
+				if apierrors.IsAlreadyExists(err) {
+					return ctrl.Result{Requeue: true}, nil
```
danielvegamyhre, line 113: Nit: can you add a comment briefly explaining this? Something like `"If the configMap was already created by a different concurrent reconcile attempt during the this window where we saw the configMap didn't exist and then tried to create it, then we re-enqueue the jobset for reconciliation so the next fetch of the configMap succeeds and the remaining logic can proceed."`

This way future readers/developers can understand this more quickly.

tpu-provisioner/internal/controller/synced_start_controller.go
```diff
@@ -0,0 +1,223 @@
...
+func startedThreshold(js *jobset.JobSet) (int32, bool) {
+	cmName := js.Annotations[jobsetSyncedStartAnnotation]
+
+	var (
+		n     int32
+		found bool
+	)
+
+	for _, rj := range js.Spec.ReplicatedJobs {
+		for _, v := range rj.Template.Spec.Template.Spec.Volumes {
```
danielvegamyhre, line 214: We already exit the main reconcile loop early if the configMap name annotation doesn't exist on the JobSet, do we need to do this check here?

Unless there's another reason for it, I think we can just do:
```
podsExpected = 0
for each replictedJob:
  podsExpected += replicatedJob.replicas * replicatedJob.template.Paralellism
```

tpu-provisioner/internal/controller/synced_start_controller.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	if !thresholdMet {
+		return ctrl.Result{}, nil
+	}
+
+	// Update the ConfigMap to indicate that all Pods have started.
+	cm.Data[allStartedConfigMapKey] = "true"
+
+	if err := ctrl.SetControllerReference(js, cm, r.Scheme); err != nil {
```
danielvegamyhre, line 165: I think we are setting the configMap controller reference twice accidentally (first time on line 103)

tpu-provisioner/internal/controller/synced_start_controller.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *SyncedStartReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&jobset.JobSet{}).
+		Watches(&corev1.Pod{}, handler.EnqueueRequestsFromMapFunc(handler.MapFunc(jobsetForPod))).
```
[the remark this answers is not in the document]
danielvegamyhre: Why is the watch necessary at all though? Can't we just reconcile on JobSets (and re-enqueue if threshold not met) without handling all these duplicate requests?

tpu-provisioner/internal/controller/synced_start_controller.go
nstogner: The thought here is that the user might have created the configmap and the controller would adopt it. So I set the reference before an update in case the create-path was never taken. It confused me re-reading it too, might try to refactor to something cleaner.

tpu-provisioner/internal/controller/synced_start_controller.go
nstogner: I was thinking that there might be replicated jobs that are not a part of the synced-start set of pods.

danielvegamyhre: I see, sounds good

tpu-provisioner/internal/controller/synced_start_controller.go
danielvegamyhre: I see. If the configMap already has the owner reference, will this line return an error and prevent the Update() call below from being executed?

tpu-provisioner/internal/controller/synced_start_controller.go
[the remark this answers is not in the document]
nstogner: The controller manager already has a watch on Pods, but yes the work queue, it is definitely doing some additional work to de-dupe here. I will look at switching to polling.

tpu-provisioner/internal/controller/pods_started_controller.go
nstogner: Resolved by reconciling on JobSets and polling until threshold is met.

tpu-provisioner/internal/controller/pods_started_controller.go
nstogner: Completed

tpu-provisioner/cmd/main.go
```diff
@@ -109,30 +110,22 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Scheme:                 scheme,
-		MetricsBindAddress:     metricsAddr,
-		Port:                   9443,
+		Scheme: scheme,
+		Metrics: metricsserver.Options{
```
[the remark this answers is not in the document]
nstogner: Yes

tpu-provisioner/internal/controller/creation_controller.go
```diff
@@ -96,5 +99,8 @@ func (r *CreationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 func (r *CreationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Pod{}).
+		WithOptions(controller.Options{
+			MaxConcurrentReconciles: r.Concurrency,
```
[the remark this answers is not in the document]
nstogner: We were doing it in main, but I moved it here after updating controller runtime (fields were moved around)

tpu-provisioner/internal/controller/synced_start_controller.go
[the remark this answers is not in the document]
nstogner: Updated to poll at 1s

tpu-provisioner/internal/controller/synced_start_controller.go
nstogner: I removed this line: I think if the user created the CM they are responsible for deleting it. If the controller created it - it will assign the JobSet as the owner. Thoughts?

danielvegamyhre: Sounds good to me

danielvegamyhre: One more thing, I notice the configMap name is hardcoded, so we are using the same configMap for all jobsets? Shouldn't we create 1 configMap per JobSet?

tpu-provisioner/cmd/main.go
```diff
@@ -61,6 +61,7 @@ var (
 
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(jobset.AddToScheme(scheme))
```
nstogner, line 64: Comment for link

spencer-p:
@nstogner Do you have someone to review this PR? Also, please rebase if this PR is still active.

nstogner:
This is abandoned at the moment, thanks for checking in

PR #564, opened 2024-04-04
Update README.md

artemvmin:
Remove instructions for creating a cluster with terraform. This is already done for you as part of the rag application.

### DIFF
--- a/applications/rag/README.md
+++ b/applications/rag/README.md
@@ -29,22 +29,6 @@ Install the following on your computer:
 * [Helm](https://helm.sh/docs/intro/install/)
 * [Gcloud](https://cloud.google.com/sdk/docs/install)
 
-### Bring your own cluster (optional)
-
-By default, this tutorial creates a Standard cluster on your behalf. We highly recommend following the default settings.
-
-If you prefer to manage your own cluster, set `create_cluster = false` in the [Installation section](#installation). Creating a long-running cluster may be better for development, allowing you to iterate on Terraform components without recreating the cluster every time.
-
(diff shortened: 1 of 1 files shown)


applications/rag/README.md
```diff
@@ -29,22 +29,6 @@ Install the following on your computer:
 * [Helm](https://helm.sh/docs/intro/install/)
 * [Gcloud](https://cloud.google.com/sdk/docs/install)
 
-### Bring your own cluster (optional)
```
andrewsykim, line 32: I think it's fine to keep this because it's marked optional. The terraform route should be flexible enough to support this option for really advanced users IMO. But I think it can be even clearer that most users should use new clusters instead

applications/rag/README.md
```diff
@@ -29,22 +29,6 @@ Install the following on your computer:
...
-
-1. `cd ai-on-gke/infrastructure`
-
-2. Edit `platform.tfvars` to set your project ID, location and cluster name. The other fields are optional. Ensure you create an L4 nodepool as this tutorial requires it.
-
-3. Run `terraform init`
-
-4. Run `terraform apply --var-file workloads.tfvars`
-
 ### Bring your own VPC (optional)
```
andrewsykim, line 32: Change this part too if you're changing the "Bring your own cluster" section?

applications/rag/README.md
imreddy13: Yeah, I want to keep this. I'm ok to simplify to a gcloud step once we switch back to Autopilot.

PR #567, opened 2024-04-04
[DO NOT MERGE] Remove GCP restricted regions from regions list

imreddy13:


artemvmin (dismissed):

artemvmin (commented):
These regions look fine according to https://cloud.google.com/compute/docs/regions-zones

PR #569, opened 2024-04-05
set enable_gpu=false in Ray workloads.tfvars file

andrewsykim:
This doesn't change any behavior since we already default `enable_gpu` to false. However, we should still specify it in the defauklt `workloads.tfvars` field for better discoverabiltiy.  

Also clean up some references and guides for Ray


applications/ray/workloads.tfvars
```diff
@@ -40,6 +40,7 @@ gcs_bucket                      = "ray-bucket-zydg"
 create_ray_cluster              = true
 ray_cluster_name                = "ray-cluster"
 enable_grafana_on_ray_dashboard = false
+enable_gpu                      = false
```
andrewsykim, line 43: This is already defaulting to false, just adding for visibility

richardsliu (approved):

Issue #509, opened 2024-03-30
Enable Image Streaming by default


andrewsykim:
We use modules in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine for creating GKE clusters, which does not expose a config for image streaming yet. 

We should either propose a fix to allow clusters to enable image streaming with that module or remvoe our depdendency to that module.

andrewsykim:
Looks like our RAG solution actually enables image streaming already which is great :) 

For other deployments we can still enable it at the node pool level by setting `enable_gcfs=true`

andrewsykim:
Might also explain flakes when pulling Ray images: https://github.com/GoogleCloudPlatform/ai-on-gke/issues/520

Issue #520, opened 2024-04-02
CI flakes sometimes when waiting to pull Ray images


andrewsykim:
Ray images (especially with GPU support) are very large and can take several minutes to pull. This has resulted in some CI checks failing while waiting for the Ray cluster to be ready.

andrewsykim:
I'm not sure why this was closed by the other PR since the flake is not related to stockouts. I also saw this again in https://github.com/GoogleCloudPlatform/ai-on-gke/pull/606

fcabrera23:
This guide was migrated to https://github.com/ai-on-gke/quick-start-guides - If you still have this issue, please open a new Issue in that new repository.

PR #570, opened 2024-04-05
Enable image streaming by default for CPU pools

andrewsykim:
We currently enable image streamign for GPU node pools but not CPU node pools. Not sure why we only enabled it for GPU node pools but it probably makes sense to enable for all node pools by default. 

Fixes https://github.com/GoogleCloudPlatform/ai-on-gke/issues/509


andrewsykim:
Might explain the flakes we're seeing when pulling Ray images: https://github.com/GoogleCloudPlatform/ai-on-gke/issues/520

imreddy13 (approved):

PR #578, opened 2024-04-10
Add Webhook Unit Test Suite

ryanaoleary:
This PR adds a comprehensive unit test suite for the Ray TPU webhook. These unit tests can be run with the command `go test` from the `applications/ray/kuberay-tpu-webhook` folder. These tests cover 75.3% of statements in the webhook including all major pieces of logic.

This PR also updates some of the logic in the webhook code to resolve corner cases found during testing. This includes adding a `getNumTPUChipsRequested` function which returns the `google.com/tpu` request value of a container, using that value to calculate the # TPU VM hosts for a given TPU worker group (rather than using `acceleratorType` to guess the `machineType`). This PR also removes the `headlessServiceName` var and instead dynamically generates the headless service name for any given cluster.

A follow-up PR will add end-to-end tests for the webhook.

This PR has been tested with:
- [x] Unit tests
- [x] Manual tests


applications/ray/kuberay-tpu-webhook/Troubleshooting.md
```diff
@@ -0,0 +1,44 @@
+# Ray TPU Webhook Troubleshooting / FAQ Guide
+Common issues and their solutions when deploying Ray TPU worker groups with the webhook.
+Solutions will be added as new issues are encountered.
+
+## `TPU_WORKER_HOSTNAMES` aren't injected into the Pod environment
```
richardsliu, line 5: What are the symptoms of this? It should start with some user error, like some JAX initialization error.

Also to get here, the user should first do a `kubectl describe` and check if the environment variables exist.

applications/ray/kuberay-tpu-webhook/Troubleshooting.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Solution #1
+`TPU_WORKER_HOSTNAMES` are only injected for multi-host worker groups. If you're expecting `TPU_WORKER_HOSTNAMES` to be injected, check that the `NumOfHosts` field in your Ray worker group spec is set to a value greater than 1.
```
richardsliu, line 8: I think there is another corner case where the TPU_WORKER_IDs could be incorrectly assigned if a delete/recreate with the same Ray cluster name happens too quickly. Maybe suggest creating clusters with unique names or wait a few seconds before retrying.

applications/ray/kuberay-tpu-webhook/Troubleshooting.md
ryanaoleary: Done, added a "Symptoms" section for this issue that describes the error you might encounter and how to check for it.

applications/ray/kuberay-tpu-webhook/Troubleshooting.md
ryanaoleary: Done.

applications/ray/kuberay-tpu-webhook/Troubleshooting.md
```diff
@@ -0,0 +1,52 @@
+# Ray TPU Webhook Troubleshooting / FAQ Guide
+Common issues and their solutions when deploying Ray TPU worker groups with the webhook.
+Solutions will be added as new issues are encountered.
+
+## `TPU_WORKER_HOSTNAMES` aren't injected into the Pod environment
+
+### Symptoms
+This may be the issue if multi-host Jax initialization fails with error `RuntimeError: Unable to initialize backend 'tpu': UNKNOWN: TPU initialization failed`. Verify that `TPU_WORKER_HOSTNAMES` are missing from the Pod environment with `kubectl describe`.
```
richardsliu, line 8: Post a sample of what to check when running `kubectl describe`?

ryanaoleary: Done.

richardsliu (approved):

PR #580, opened 2024-04-10
Cherrypick AP cloud build stockout mitigation onto release-1.1

artemvmin:
Remove timeout from RayCluster CR apply; bump CB timeout to mitigate stockouts (#576)

RayCluster apply takes O(seconds). The actual ray worker deployment is done asynchronously by the ray operator.

### DIFF
--- a/modules/kuberay-cluster/main.tf
+++ b/modules/kuberay-cluster/main.tf
@@ -34,9 +34,6 @@ resource "helm_release" "ray-cluster" {
   namespace        = var.namespace
   create_namespace = true
   version          = "1.0.0"
-  # Timeout is increased to guarantee sufficient scale-up time for Autopilot nodes.
-  timeout = 1200
-  wait    = true
 
   values = [
     templatefile("${path.module}/values.yaml", {
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -99,7 +99,7 @@ steps:
         echo "pass" > /workspace/user_result.txt
 
         # Make sure pods are running
-        kubectl wait --all pods -n ml-$SHORT_SHA-$_BUILD_ID --for=condition=Ready --timeout=300s
+        kubectl wait --all pods -n ml-$SHORT_SHA-$_BUILD_ID --for=condition=Ready --timeout=1200s
         kubectl port-forward -n ml-$SHORT_SHA-$_BUILD_ID service/ray-cluster-kuberay-head-svc 8265:8265 &
         # Wait port-forwarding to take its place
         sleep 5s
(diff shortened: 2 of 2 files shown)


modules/kuberay-cluster/main.tf
```diff
@@ -34,9 +34,6 @@ resource "helm_release" "ray-cluster" {
   namespace        = var.namespace
   create_namespace = true
   version          = "1.0.0"
-  # Timeout is increased to guarantee sufficient scale-up time for Autopilot nodes.
-  timeout = 1200
```
andrewsykim, line 38: From the PR description:

>"RayCluster apply takes O(seconds). The actual ray worker deployment is done asynchronously by the ray operator."

That means this terraform resource currently completes quickly and the 1200 timeout was not actually waiting for scale up right? Wondering how removing it helps with stockout mitigations.

andrewsykim: Nvm I think I answered my own questions. It's because wait = true is waiting for resources to be ready

andrewsykim: Does this mean in a stockout scenario we can have a deployment pass while the Ray cluster is still not ready?

artemvmin: This module only deploys the custom resource, not the pods themselves. It always completes in ~5 seconds, so this timeout isn't doing anything. We need to wait asynchronously for the kuberay-operator to spin up the deployment in response to the custom resource deployed by this module.

andrewsykim: Thanks for clarifying -- that makes sense. How does just removing the timeout (that doesn't do anything already) help with stockout mitigation?

imreddy13: This is true, and also the timeout increase to 20 mins helping with CI stability makes me think this isn't actually a stockout but cluster autoscaler taking a while to scale up the nodes. In my experience, stockouts take a lot longer to resolve. Wdyt @artemvmin ?

artemvmin: The continuous build for AP has been at 100% success rate since we checked this change into main branch.

> The cloudbuild change makes sense to me, but that's only applicable in our tests. I'm not seeing how removal of the timeout helps with stockout mitigation for users trying the quick starts.

The test enforces an artificial 5-minute deadline which isn't present during actual usage.

> the timeout increase to 20 mins helping with CI stability in the main branch makes me think this isn't actually a stockout but cluster autoscaler taking a while to scale up the nodes

I haven't seen any evidence of Autoscaler having issues, while stockouts are quite common. The difference between the 5-minute cloud build timeout and the 30-minute node pool creation timeout explains why Standard runs used to appear more stable.

andrewsykim: > The continuous build for AP has been at 100% success rate since we checked this change into main branch.

This makes sense, because the change only fixes cloudbuild run by increasing the timeout when running `kubectl wait`. But actual execution of the terraform module does not have this wait time at all. 

Just trying to understand how this specific change helps with stockout mitigation outside of the cloudbuild run

artemvmin: As you said, this change doesn't fix stockouts; it makes our tests less flaky. The Standard tests get around this by putting a 35+ minute wait earlier in the cloud build script (during node pool creation).

andrewsykim: Oh okay, I was confused with the PR title and description that says this mitigates stockouts.

artemvmin: Updated title

andrewsykim (approved):

PR #585, opened 2024-04-12
Update webhook cert-manager install to work with GKE autopilot clusters

ryanaoleary:
This PR changes the `make install-cert-manager` command to one that works for both GKE standard and autopilot clusters. The previous `kubectl apply` command to install cert-manager is incompatible with GKE autopilot since the default `global.leaderElection.namespace` is a managed namespace on autopilot.

 This PR was tested by installing cert-manager using the helm commands on both standard and autopilot clusters and verifying that the webhook was successfully issued a certificate.

### DIFF
--- a/applications/ray/kuberay-tpu-webhook/Makefile
+++ b/applications/ray/kuberay-tpu-webhook/Makefile
@@ -31,16 +31,7 @@ docker-build:
   
 # Push the docker image  
 docker-push:  
-	docker push ${IMG}  
-
-# Install once per cluster
-install-cert-manager:  
-	helm repo add jetstack https://charts.jetstack.io
-	helm repo update
-	helm install --create-namespace --namespace cert-manager --set installCRDs=true --set global.leaderElection.namespace=cert-manager cert-manager jetstack/cert-manager
-
-uninstall-cert-manager:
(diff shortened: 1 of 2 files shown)


applications/ray/kuberay-tpu-webhook/Makefile
```diff
@@ -32,12 +32,15 @@ docker-build:
 # Push the docker image  
 docker-push:  
 	docker push ${IMG}  
-  
+
+# Install once per cluster
 install-cert-manager:  
-	kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.13.2/cert-manager.yaml
+	helm repo add jetstack https://charts.jetstack.io
```
richardsliu, line 38: Maybe we should explicitly document this as a prerequisite in the user guide instead of here. Also add explanations for why the user needs to install their own cert manager.

ryanaoleary: I'll add it as a pre-requisite for each cluster and an explanation in the TPU_GUIDE. Should I remove the `make` commands then and just leave the install commands in the guide?

richardsliu (approved):

PR #587, opened 2024-04-15
Adding Jetstream support to locust solution.

kfswain:
In addition to Jetstream support:

- Made --stop-timeout an opt-in variable
- Piped namespace var to locust master KNS name
- Bumped transformers version for Gemma support

### DIFF
--- a/benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/run.sh
+++ b/benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/run.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 LOCUST="/usr/local/bin/locust"
-LOCUS_OPTS="-f /locust-tasks/tasks.py --host=$TARGET_HOST"
+LOCUST_OPTS="-f /locust-tasks/tasks.py --host=$TARGET_HOST"
 LOCUST_MODE=${LOCUST_MODE:-standalone}
 
 if [[ "$LOCUST_MODE" = "master" ]]; then
@@ -24,7 +24,10 @@ if [[ "$LOCUST_MODE" = "master" ]]; then
     # For inferencing workloads with large payload having no wait time is unreasonable.
     # This timeout is set to large amount to avoid user tasks being killed too early.
     # TODO: turn timeout into a variable.
-    LOCUS_OPTS="$LOCUS_OPTS --master --stop-timeout 10800"
+    LOCUST_OPTS="$LOCUST_OPTS --master "
+    if [[ "$STOP_TIMEOUT" != 0 ]]; then
+        LOCUST_OPTS="$LOCUST_OPTS --stop-timeout $STOP_TIMEOUT"
+    fi
 elif [[ "$LOCUST_MODE" = "worker" ]]; then
     huggingface-cli login --token $HUGGINGFACE_TOKEN
     FILTER_PROMPTS="python /locust-tasks/load_data.py"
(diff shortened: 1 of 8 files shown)


benchmarks/benchmark/tools/locust-load-inference/variables.tf
```diff
@@ -71,6 +71,12 @@ variable "num_locust_workers" {
   default     = 1
 }
 
+variable "stop_timeout" {
+  description = "Length of time before a locust job is stopped."
+  type        = number
+  default     = 0
```
annapendleton, line 77: Default should be 10800 to keep consistent behavior. Or is 0 a more reasonable default in this case?

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/run.sh
```diff
@@ -24,17 +24,20 @@ if [[ "$LOCUST_MODE" = "master" ]]; then
     # For inferencing workloads with large payload having no wait time is unreasonable.
     # This timeout is set to large amount to avoid user tasks being killed too early.
     # TODO: turn timeout into a variable.
-    LOCUS_OPTS="$LOCUS_OPTS --master --stop-timeout 10800"
+    LOCUST_OPTS="$LOCUST_OPTS --master "
+    if [[ "$STOP_TIMEOUT" != 0 ]]; then
+        LOCUST_OPTS="$LOCUST_OPTS --stop-timeout $STOP_TIMEOUT"
```
annapendleton, line 29: OOC, why does this need to be an opt-in var? Is this related to making the runner work?

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
```diff
@@ -123,6 +128,8 @@ def get_token_count(prompt, resp):
             tokenizer.encode(resp_dict['text_output']))
     elif backend == "sax":
         number_of_output_tokens = 0  # to be added
+    elif backend == "jetstream":
+        number_of_output_tokens = 0
```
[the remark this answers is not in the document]
kfswain: Yeah, I was just adding barebones support and fixes. Will follow up with a metrics-related PR

benchmarks/benchmark/tools/locust-load-inference/variables.tf
kfswain: I chose `0`, as that forces this flag to be opt-in. My thinking is making the field opt-in reduces the chance of hitting: https://github.com/locustio/locust/issues/2621. Which I hit during development

LMKWYT

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/run.sh
kfswain: Addressed in the previous comment

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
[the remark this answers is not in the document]
annapendleton: Sounds good ty for clarifying!

benchmarks/benchmark/tools/locust-load-inference/variables.tf
annapendleton: I tested locally with 0 with my current tests and confirmed it has no negative impact. Also looks like stop_timeout doesn't do anything in terms of stopping long running user tasks as I had thought - it actually impacts "graceful termination" - https://github.com/locustio/locust/blob/a2f328fdb2fcf5592b07b737680a9edcfa201301/locust/env.py#L83

annapendleton: With that - I'm fine with this default set to 0 :) thanks for explaining and fixing this!

annapendleton (approved):

PR #590, opened 2024-04-18
Delete node pools for Nodes in an unknown state for too long

nstogner:


tpu-provisioner/internal/controller/deletion_controller.go
```diff
@@ -69,8 +69,21 @@ func (r *DeletionReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
+	if !ok {
+		lg.V(3).Info("No node pool label found on node, ignoring", "labelKey", nodePoolLabelKey)
+		return ctrl.Result{}, nil
+	}
+
 	// Avoid noisy reconciliation when nodes are shutting down.
 	for _, c := range node.Status.Conditions {
+		if c.Type == corev1.NodeReady &&
+			c.Status == corev1.ConditionUnknown &&
+			time.Since(c.LastTransitionTime.Time) > 20*time.Minute {
```
thebinaryone1, line 83: Can we change it to 10 minutes instead?

thebinaryone1 (approved):

tpu-provisioner/internal/controller/deletion_controller.go
danielvegamyhre: I'll do this in a follow up PR

roberthbailey:
Is this still under development or should we close this as obsolete?

nstogner:
Ahh, this is from so long ago that I cant remember. We can close for now. I will reopen if needed.

PR #597, opened 2024-04-22
Change nodepool name to a placeholder

pajakd:
We are including this example in tutorials (https://cloud.devsite.corp.google.com/kubernetes-engine/docs/how-to/provisioningrequest#run_your_job). To enable the node pool name to be displayed as an editable variable in the tutorial I need to reformat it here.

Node pool is not defined in this example (it is only referenced) so the user might want to use any name. Having a placeholder makes it clearer.


andrewsykim (approved):
Defer lgtm to @alculquicondor

PR #601, opened 2024-04-22
Add event filters for TPU Provisioner creation controller

danielvegamyhre:
This change ensures the TPU Provisioner creation controller only reconciles pods matching the following conditions:

- Part of a JobSet
- Leader pod (job completion index 0)
- Not yet scheduled
- Not marked for deletion
- Pod does not have label `tpu-provisioner.cloud.google.com/disable-autoprovisioning=true`

### DIFF
--- a/tpu-provisioner/internal/controller/creation_controller.go
+++ b/tpu-provisioner/internal/controller/creation_controller.go
@@ -52,6 +52,7 @@ type PodCriteria struct {
 //+kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups="",resources=pods/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups="",resources=pods/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
 
 func (r *CreationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	lg := ctrllog.FromContext(ctx)
@@ -98,6 +99,7 @@ func (r *CreationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				isUnschedulable(pod) &&
 				doesRequestResource(pod, r.PodCriteria.ResourceType) &&
 				hasNodeSelectors(pod, cloud.GKETPUNodeSelector) &&
+				!autoProvisioningDisabled(pod) &&
 				!podDeleted(pod)
 		})).
 		Complete(r)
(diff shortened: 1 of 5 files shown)


tpu-provisioner/internal/controller/creation_controller.go
```diff
@@ -96,5 +98,34 @@ func (r *CreationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 func (r *CreationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Pod{}).
+		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
+			// Only reconcile JobSet leader pods which have not been scheduled and
+			// have not been marked for deletion.
+			pod, ok := object.(*corev1.Pod)
+			return ok && podPartOfJobSet(pod) && isLeaderPod(pod) && !podScheduled(pod) && !podDeleted(pod)
```
nstogner, line 105: Can you remove the similar checks from the Reconcile() func and move all to here and add something into the integration test that checks it is still working?

nstogner (changes requested):

tpu-provisioner/internal/controller/creation_controller.go
danielvegamyhre: Consolidated pod filtering logic in the event filter and updated integration test with more cases. I also refactored the integration test structure to be more organized and allow us to easily add more cases in the future.

tpu-provisioner/internal/controller/creation_controller.go
```diff
@@ -96,5 +88,17 @@ func (r *CreationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 func (r *CreationReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Pod{}).
+		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
+			// Only reconcile pods which meet the conditions defined below.
+			pod, ok := object.(*corev1.Pod)
+			return ok &&
```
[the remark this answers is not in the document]
danielvegamyhre: Sure, added a check for this and a new test case for it as well.

Also introduced a constants.go file to the controller and moved kubebuilder marker for events RBAC to the creation controller file instead of the empty events.go file (renamed to constants.go).

nstogner (approved):
LGTM - good improvements to tests!

thebinaryone1 (approved):

PR #603, opened 2024-04-22
Move Ray TPU Webhook out of applications/ray folder

ryanaoleary:
This PR moves the Ray TPU webhook from the `applications/ray` folder (reserved for terraform templates) to a new `ray-on-gke/tpu` folder. This PR contains no changes to the webhook code.


andrewsykim:
I've made the presubmits optional until we resolve the HF issue

andrewsykim (approved):

andrewsykim:
I'm skipping the cloudbuild presubmit since we don't test kuberay TPU webhook there

PR #604, opened 2024-04-23
Add a troubleshoot section for the gated Mistral-7B-Instruct model issue

gongmax:


jsoohoo-google:
Update docs for temporary HF model gating workaround.

PR #605, opened 2024-04-23
Bump rag frontend resource requests

artemvmin:
Frontend sometimes gets OOM-killed. Set guaranteed QOS.


blackzlq (approved):

blackzlq:
The dangerous part is https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/rag/frontend/container/cloud_sql/cloud_sql.py#L58 we select all. LOL, I think Julie's change will reduce this issue.

PR #608, opened 2024-04-24
Update Webhook Release Version Schema

ryanaoleary:
This PR adds a VERSION file to the Ray TPU webhook folder in order to follow [semantic versioning](https://semver.org/). This file will contain the version of the webhook following the format `v(MAJOR).(MINOR).(PATCH)`. This version number is separate from the release version of the `ai-on-gke` repo. This PR also changes the Docker image path to a folder where new images are automatically generated by commits to the webhook folder. These images use the string specified in VERSION for the Docker image tag.

Any future commits to the webhook should increment the value in the VERSION file, changing the `image:` value in `deployments/deployment.yaml` to match.


applications/ray/kuberay-tpu-webhook/VERSION
```diff
@@ -0,0 +1,4 @@
+# Release version of the Ray TPU webhook
+# Increment this value as part of any change to the webhook.
+# This value is separate from the ai-on-gke release version.
+v1.1.0
```
andrewsykim, line 4: I thought we were using tags instead? Keeping track of a version in a file will be difficult to track and easy to override existing versions

ryanaoleary: That makes sense, I was just having trouble finding a way to create a git tag that would apply to just the webhook. I talked to Spencer and he said it was alright to create `kuberay-tpu-webhook/vX.Y.Z` tags that apply to the whole repo, so I'll change it to use that instead of the VERSION file.

PR #609, opened 2024-04-24
Update Dockerfile transformer install order

thisSIDEofRANDOM:
Swapping the checkout and install to make sure we install the checkout and not the latest transformers


thisSIDEofRANDOM:
@ahg-g PTAL

andrewsykim:
FYI: there is a known bug that is causing presubmits to fail. They are optional for now and can be ignored

ahg-g (approved):

PR #614, opened 2024-04-25
TPU Provisioner reliability improvements

danielvegamyhre:
This PR includes the following changes:

- Migrate from Job UID to JobSet job-key label for node pool name. This will ensure if 1 of N node pool is in a bad state, every time the JobSet restarts and a different child Job's pods become unschedulable, this won't trigger a NP recreation since the Job UID (which changes between restarts) is no longer used in the node pool name.

- Track name of JobSet whose pods triggered the creation of a given node pool. In the node pool garbage collection loop, use this information to determine if we should delete a node pool which is an ERROR state with no k8s node objects associated with it, whose JobSet no longer exists.

- Upgrade to controller-runtime and other packages to latest versions. This was require to use import JobSet API package. The previous controller-runtime version was pretty old (v0.14.1 v latest v0.18.0) so there were some breaking changes I had to resolve.


tpu-provisioner/internal/cloud/gke.go
```diff
@@ -57,7 +63,7 @@ type GKE struct {
 func (g *GKE) NodePoolLabelKey() string { return GKENodePoolNameLabel }
 
 func (g *GKE) EnsureNodePoolForPod(p *corev1.Pod, why string) error {
-	name := podToNodePoolName(p, GKENodePoolNamePrefix, "")
+	name, err := podToNodePoolName(p)
```
nstogner, line 66: Unchecked error

tpu-provisioner/internal/cloud/gke.go
```diff
@@ -340,23 +352,24 @@ func sumTPURequests(p *corev1.Pod) (int, error) {
...
+func podToNodePoolName(p *corev1.Pod) (string, error) {
+	jobSetName, exists := p.Labels[jobset.JobSetNameKey]
+	if !exists {
+		return "", fmt.Errorf("%s label not found on pod %s", jobset.JobSetNameKey, p.Name)
+	}
+	jobKey, exists := p.Labels[jobset.JobKey]
+	if !exists {
+		return "", fmt.Errorf("%s label not found on pod %s", jobset.JobKey, p.Name)
+	}
+	nodePoolName := fmt.Sprintf("%s-%s", jobSetName[:34], jobKey[:5])
```
nstogner, line 371: `:34` will panic if the name is not that long

nstogner (changes requested):

tpu-provisioner/internal/cloud/gke.go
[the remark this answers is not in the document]
danielvegamyhre: Woops - fixed.

tpu-provisioner/internal/cloud/gke.go
danielvegamyhre: Ah I was assuming slicing in Go worked the same way as in Python, I guess not! Thanks, fixed.

tpu-provisioner/internal/controller/nodepool_garbage_collector.go
```diff
@@ -46,25 +47,25 @@ func (g *NodePoolGarbageCollector) Run(ctx context.Context) {
 		for _, np := range nodepools {
 			log := log.WithValues(
 				"nodepool", np.Name,
-				"createdForPodName", np.CreatedForPod.Name,
-				"createdForPodNamespace", np.CreatedForPod.Namespace,
+				"CreatedForJobSetName", np.CreatedForJobSet.Name,
```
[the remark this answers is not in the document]
danielvegamyhre: Done

nstogner (approved):
LGTM, we could add a check to see if the JobSet is still active or not as well as existent - or wait to add in another PR

danielvegamyhre:
> LGTM, we could add a check to see if the JobSet is still active or not as well as existent - or wait to add in another PR

Makes sense, I have the jobset utils for checking status etc included as part of the follow up PR for the deletion controller update. I'll include that change in that one.

thebinaryone1 (approved):

tpu-provisioner/internal/controller/deletion_controller.go
```diff
@@ -183,12 +184,15 @@ func (r *DeletionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return fmt.Errorf("NodeCriteria.MinLifetime must be set")
 	}
 
-	// NOTE: Direct Node watches are filtered based on labels in main.go.
-	//       However, Reconcile() can still be called for non-filtered Nodes
-	//       Because the of Watch on Pods below.
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Node{}).
-		Watches(&source.Kind{Type: &corev1.Pod{}}, handler.EnqueueRequestsFromMapFunc(handler.MapFunc(nodeForPod))).
```
nstogner, line 191: This might cause issues: deletions might take a long time b/c reconcile might not be called when the Pods leave the Nodes.

PR #615, opened 2024-04-25
Pass ICI Resiliency label to Node Pool Creation Request

SidneyShen:
This change will pass the ICI resiliency label (cloud.google.com/gke-tpu-ici-resiliency) to node pool creation request if pod has such node selector specified.

### DIFF
--- a/tpu-provisioner/internal/cloud/gke.go
+++ b/tpu-provisioner/internal/cloud/gke.go
@@ -28,6 +28,7 @@ const (
 	GKETPUNodeSelector         = "cloud.google.com/gke-tpu-topology"
 	GKEAcceleratorNodeSelector = "cloud.google.com/gke-tpu-accelerator"
 	GKENodePoolNameLabel       = "cloud.google.com/gke-nodepool"
+	ICIResiliencyLabel         = "cloud.google.com/gke-tpu-ici-resiliency"
 
 	// Supported accelerator types
 	V4PodSliceAccelerator  = "tpu-v4-podslice"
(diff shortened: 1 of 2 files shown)


SidneyShen:
@danielvegamyhre @nstogner, pls review.

tpu-provisioner/internal/cloud/gke.go
```diff
@@ -33,6 +33,7 @@ const (
 	GoogleTPUResource          = "google.com/tpu"
 	gcpLabelPrefix             = "cloud.google.com/"
 	googleLabelPrefix          = "google.com/"
+	ICIResiliencyLabel         = "cloud.google.com/gke-tpu-ici-resiliency"
```
nstogner, line 36: Does the label control ici-resiliency or is the label applied based on an option supplied when creating the node pool?

SidneyShen: The latter, we are using the labels from node selector to instruct provisioner or NAP to create the node pools. GKE control plane will interpret this info and pass the intent via VM metadata to GCE.

nstogner (approved):

tpu-provisioner/internal/cloud/gke_test.go
```diff
@@ -128,3 +135,208 @@ func Test_tpuMachineType(t *testing.T) {
...
+						{
+							APIVersion:         "batch/v1",
+							Kind:               "Job",
+							UID:                "8484279a-de52-4ca1-b01e-130fbded30fb",
+							Name:               "jobset-test-job-1-0",
+							Controller:         &trueVar,
+							BlockOwnerDeletion: &trueVar,
+						},
+					},
+					GenerateName:    "jobset-test-job-1-0-0-",
```
danielvegamyhre, line 305: Why are both Name and GenerateName set? Shouldn't it be one or the other?

danielvegamyhre (commented):
LGTM after rebase, left one minor comment/question

tpu-provisioner/internal/cloud/gke_test.go
SidneyShen: Test locally, both are set, the generated name should be set by job contoller.

danielvegamyhre (approved):

PR #617, opened 2024-04-25
Fixed redirect

arueth:


alizaidis (approved):
LGTM

achandrasekar:
Would be better to create a symlink like we have for `jupyter-on-gke` in order to not show up as separate directory at the top of the list. @imreddy13 thoughts?

alizaidis:
This PR was intended to fix the behavior of users receiving a 404 when you typed in Google "gke batch reference architecture" and clicked on the first result which points to the out dated readme path.

PR #618, opened 2024-04-26
Set the default GKE cluster type for ray to GKE Autopilot.

roberthbailey:
Also add instructions to use a standard cluster if preferred.

### DIFF
--- a/applications/ray/README.md
+++ b/applications/ray/README.md
@@ -3,22 +3,58 @@
 This repository contains a Terraform template for running [Ray](https://www.ray.io/) on Google Kubernetes Engine.
 See the [Ray on GKE](/ray-on-gke/) directory to see additional guides and references.
 
+## Prerequisites
+
+1. GCP Project with following APIs enabled
+    - container.googleapis.com
+    - iap.googleapis.com (required when using authentication with Identity Aware Proxy)
+
+2. A functional GKE cluster.
+    - To create a new standard or autopilot cluster, follow the instructions in [`infrastructure/README.md`](https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/infrastructure/README.md)
+    - Alternatively, you can set the `create_cluster` variable to true in `workloads.tfvars` to provision a new GKE cluster. This will default to creating a GKE Autopilot cluster; if you want to provision a standard cluster you must also set `autopilot_cluster` to false.
+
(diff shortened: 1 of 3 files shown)


applications/ray/README.md
```diff
@@ -3,22 +3,58 @@
 This repository contains a Terraform template for running [Ray](https://www.ray.io/) on Google Kubernetes Engine.
 See the [Ray on GKE](/ray-on-gke/) directory to see additional guides and references.
 
+## Prerequisites
+
+1. GCP Project with following APIs enabled
```
jsoohoo-google, line 8: Can we enable these by default?

applications/ray/variables.tf
```diff
@@ -39,7 +39,7 @@ variable "ray_version" {
 variable "kubernetes_namespace" {
   type        = string
   description = "Kubernetes namespace where resources are deployed"
-  default     = "myray"
+  default     = "ml"
```
jsoohoo-google, line 42: Is this really the right namespace? Should it be gke-ai-ray or something more meaningful?

applications/ray/README.md
roberthbailey: I copied these pre-reqs from the jupyter readme. We do enable services in the terraform [here](https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/ray/main.tf#L39) so maybe this shouldn't be part of either README. Are you ok leaving it in for now?

applications/ray/variables.tf
roberthbailey: This is the namespace name that we set by default in [workloads.tfvars](https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/ray/workloads.tfvars#L30). I updated this one so that they would be consistent.

applications/ray/README.md
jsoohoo-google: ok

jsoohoo-google (approved):

applications/ray/variables.tf
roberthbailey: Also, the jupyter solution uses `ml` as well: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/jupyter/workloads.tfvars#L35

roberthbailey: And the rag solution uses `rag`: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/rag/workloads.tfvars#L27 & https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/rag/variables.tf#L42

PR #624, opened 2024-04-26
Add private_cluster_configuration instead of enable_private_endpoint.…

hubatish:
Purpose is to allow creating a public cluster just setting variables.

Tested with first a terraform apply setting `private_cluster_configuration = null` that created a public cluster, then swapping the value to {} which resulted in this terraform plan: https://imgur.com/a/PDy1roA

### DIFF
--- a/benchmarks/infra/stage-1/README.md
+++ b/benchmarks/infra/stage-1/README.md
@@ -68,27 +68,60 @@ gcloud container fleet memberships get-credentials <cluster-name>
 kubectl get nodes
 ```
 
-<!-- BEGIN TFDOC -->
-## Variables
+<!-- BEGIN_TF_DOCS -->
+Copyright 2024 Google LLC
 
-| name | description | type | required | default |
-|---|---|:---:|:---:|:---:|
-| [cluster_name](variables.tf#L22) | Name of new or existing cluster. | <code>string</code> | ✓ |  |
-| [project_id](variables.tf#L17) | Project id of existing or created project. | <code>string</code> | ✓ |  |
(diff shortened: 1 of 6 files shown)


benchmarks/infra/stage-1/variables.tf
```diff
@@ -36,10 +36,14 @@ variable "gke_location" {
   default     = "us-central1-a"
 }
 
-variable "enable_private_endpoint" {
-  description = "When true, the cluster's private endpoint is used as the cluster endpoint and access through the public endpoint is disabled."
-  type        = bool
-  default     = true
+variable "private_cluster_config" {
```
achandrasekar, line 39: It looks like we were defaulting to a private cluster before. Now, with null default, it changes it to a public cluster. Is that right?

benchmarks/infra/stage-1/main.tf
```diff
@@ -31,7 +31,7 @@ module "gke-infra" {
 
   registry_create = true
 
-  enable_private_endpoint = var.enable_private_endpoint
+  private_cluster_config = var.private_cluster_config
```
achandrasekar, line 34: README needs to be changed to reflect the variable change. Please use tfdocs to update it.00

Ideally, it would be best if this doesn't break the existing sample-terraform-vars that people are using when they don't modify it. Is that the case still?

benchmarks/infra/stage-1/variables.tf
hubatish: No, private remains the default. The default is {}, which I think then gets set to the object with variable enable_private_endpoint = true & master_global_access = true. That seems to be how it works given my two tests of {} vs null + terraform plan.

benchmarks/infra/stage-1/main.tf
hubatish: Thanks, ran `terraform-docs markdown table --output-file ./README.md` .

Unfortunately this does break the previous tfvar files for anyone who was setting `enable_private_endpoint`. That's not the case in sample-terraform.tfavrs but of course could be for anyone else. Want me to leave the old variable around & use it in the new one?

achandrasekar: Yes, let's leave the old variable around in case anyone was relying on that and use it to set the value when specified.

benchmarks/infra/stage-1/README.md
```diff
@@ -1,94 +1,485 @@
-# AI on GKE Benchmark Infrastructure
```
achandrasekar, line 1: Looks like you've replaced the entire readme. We just update the variable section with tfdocs. Not the whole readme since it includes other instructions as well. Btw this is the command
`terraform-docs markdown table --output-file README.md --output-mode inject .`.

hubatish: This is the output of that command. It does look like the delimiter changed though? From  to  . I removed the old section

achandrasekar (approved):

PR #627, opened 2024-04-29
Cherry-pick #599 and #618 to release-1.1

roberthbailey:
Cherry-pick #599 and #618 to the release-1.1 branch.

### DIFF
--- a/applications/ray/variables.tf
+++ b/applications/ray/variables.tf
@@ -39,7 +39,7 @@ variable "ray_version" {
 variable "kubernetes_namespace" {
   type        = string
   description = "Kubernetes namespace where resources are deployed"
-  default     = "myray"
+  default     = "ml"
 }
 
 variable "enable_grafana_on_ray_dashboard" {
@@ -105,7 +105,7 @@ variable "private_cluster" {
 
 variable "autopilot_cluster" {
   type    = bool
-  default = false
+  default = true
 }
 
 variable "cpu_pools" {
--- a/applications/jupyter/README.md
+++ b/applications/jupyter/README.md
@@ -20,7 +20,8 @@ This module deploys the following resources, once per user:
     - iap.googleapis.com (required when using authentication with Identity Aware Proxy)
 
 2. A functional GKE cluster.
-    - To create a new standard or autopilot cluster, follow the instructions under `infrastructure/README.md`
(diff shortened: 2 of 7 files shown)


andrewsykim (approved):

applications/ray/variables.tf
```diff
@@ -39,7 +39,7 @@ variable "ray_version" {
 variable "kubernetes_namespace" {
   type        = string
   description = "Kubernetes namespace where resources are deployed"
-  default     = "myray"
+  default     = "ml"
```
andrewsykim, line 42: can you actually make this `ai-on-gke` to be consistent with others?

andrewsykim: (not a blocking commnet)

roberthbailey: As I mentioned in #618, I set this to match the namespace name that we set in [workloads.tfvars](https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/ray/workloads.tfvars#L30) and updated this one so that they would be consistent.

roberthbailey: Also, the jupyter solution uses ml as well: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/jupyter/workloads.tfvars#L35

roberthbailey: And the rag solution uses rag: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/rag/workloads.tfvars#L27 & https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/rag/variables.tf#L42

roberthbailey: Where do you see the use of `ai-on-gke`?

roberthbailey: Discussed offline. We will follow up separately on the main branch.

andrewsykim: > I set this to match the namespace name that we set in [workloads.tfvars](https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/applications/ray/workloads.tfvars#L30) and updated this one so that they would be consistent.

IMO the default namespace in the workloads.tfvars file should also be `ai-on-gke` instead of `ml`

PR #629, opened 2024-04-29
Creating User for models that support gRPC requests. (Currently bespo…

kfswain:
Currently bespoke for JetStream. As that is the only model using gRPC atm


benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/run.sh
```diff
@@ -15,9 +15,15 @@
 # limitations under the License.
 
 LOCUST="/usr/local/bin/locust"
-LOCUST_OPTS="-f /locust-tasks/tasks.py --host=$TARGET_HOST"
+LOCUST_OPTS="-f /locust-tasks/tasks.py "
 LOCUST_MODE=${LOCUST_MODE:-standalone}
 
+if [[ "$REQUEST_TYPE" = "grpc" ]]; then
```
liurupeng, line 21: from the below change, it only assumes JetStream as the only caller, since the tasks.py takes in the JetStream proto

liurupeng:
if the grpc support is only targeting the JetStream, probably directly call it as JetStream-requester or Jet-client etc?
btw, have you tried run benchmark with the change against the JetStream server?

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
```diff
@@ -300,3 +306,78 @@ def total_content_length():
...
+    abstract = True
+    stub_class = None
+
+    def __init__(self, environment):
+        super().__init__(environment)
+        for attr_value, attr_name in ((self.host, "host"), (self.stub_class, "stub_class")):
+            if attr_value is None:
+                raise LocustError(f"You must specify the {attr_name}.")
+
+        self._channel = grpc.insecure_channel(self.host)
```
vivianrwu, line 320: Is the benchmarking done synchronously or asynchronously? (benchmarking done on the Jetstream repo uses grpc aio - https://github.com/google/JetStream/blob/main/benchmarks/benchmark_serving.py#L374

kfswain: They do. But the only reference of it is awaiting: https://github.com/google/JetStream/blob/main/benchmarks/benchmark_serving.py#L407

Also, I'm not sure we want to have asynchronous behavior for benchmarking atm. It could lead to unexpected results. I think our solution currently would be to spin up more workers.

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/run.sh
kfswain: That's right. When we seek to support other gRPC model servers/move to a generic proto. Discussed further in a different comment.

kfswain:
> if the grpc support is only targeting the JetStream, probably directly call it as JetStream-requester or Jet-client etc?
I'd like to keep as much code generic as possible to signal to any contributors to extend this code to support other gRPC Model Servers, instead of making a per server implementation. I'm hoping we can help drive towards a more generic proto since Google is the primary user of gRPC.  I'm actually going to change JetStreamUser to instead be GrpcBenchmarkUser. 

> btw, have you tried run benchmark with the change against the JetStream server?
Yes, I spun up my own cluster and JetStream implementation independently (did not use the provided stage-1/2 terraform). And then used the terraform in the locust directory to spin up the locust jobs. I can update the readme with the branching gRPC behavior

benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
```diff
@@ -45,10 +47,8 @@ spec:
...
             - name: USE_BEAM_SEARCH
               value: ${use_beam_search}
-%{ for hugging_face_token_secret in hugging_face_token_secret_list ~}
             - name: HUGGINGFACE_TOKEN
               valueFrom:
                 secretKeyRef:
-                  name: hf-token
-                  key: HF_TOKEN
-%{ endfor ~}
+                  name: huggingface-secret
```
achandrasekar, line 53: Does this change break any of the other cases?

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
```diff
@@ -300,3 +306,78 @@ def total_content_length():
...
+        exception = None
+        start_perf_counter = time.perf_counter()
+        response_length = 0
+        responses = method(request_or_iterator, call_details)
+        output = ""
+        response_length = 0
+        ttft = 0
+        for response in responses:
+            if ttft == 0:
+                ttft = time.perf_counter() - start_perf_counter
```
achandrasekar, line 369: I'm not sure I understand how ttft is being calculated here. Can you expand? Also, would be good to add a comment on what this does.

benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
kfswain: oop. Reverting. This was just for my cluster setup. A

lthough, this was changed recently and now seems to be opinionated on having the secret stored in GCP as opposed to using K8s secrets. Was this intentional?

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
kfswain: For sure! Adding a comment, but also explaining here. We stream the response, so as each chunk is sent back to us, we create the full output. When we receive our first token back, we calculate that t-delta to calculate TTFT. 

The assumption is that each chunk returned is a single token. But Jetstream makes a similar assumption: https://github.com/google/JetStream/blob/main/benchmarks/benchmark_serving.py#L381


Actual output from benchmark logs:  `response_time 5931.7820890573785; ttft:170.58807401917875`

achandrasekar: Makes sense. Thanks for the clarification!

benchmarks/benchmark/tools/locust-load-inference/README.md
```diff
@@ -110,9 +110,10 @@ Fill out your `terraform.tfvars` with the desired model and server configuration
 - `credentials_config` - credentials for cluster to deploy Locust benchmark tool on
 - `project_id` - project id for enabling dependent services for building locust artifacts
 - `artifact_registry` - artifact registry to upload locust artifacts to
-- `inference_server_service` - an accessible service name for inference workload to be benchmarked
+- `inference_server_service` - an accessible service name for inference workload to be benchmarked **(Note: a non-80 port should be specified here. Example: `my-service-name:9000`)
```
achandrasekar, line 113: Is there a reason to specify a non-80 port here? I think the other ones are creating services which are exposed in port 80 by default.

benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl
kfswain: SGTM. I'll make a note.

PR #675, opened 2024-05-20
few small fixes for http tgi route

annapendleton:
A few additional fixes required for http route

Tested end to end, validated locust comes up, and successful response.

Updates:
- remove single ' in run.sh
- remove self. references to methods that no longer belong to http user class
- update recommended defaults in sample tfvars file to work for tgi http

### DIFF
--- a/benchmarks/benchmark/tools/locust-load-inference/sample-terraform.tfvars
+++ b/benchmarks/benchmark/tools/locust-load-inference/sample-terraform.tfvars
@@ -9,7 +9,7 @@ ksa       = "benchmark-ksa"
 
 # Locust service configuration 
 artifact_registry                        = "us-central1-docker.pkg.dev/$PROJECT_ID/ai-benchmark"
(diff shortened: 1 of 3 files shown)


benchmarks/benchmark/tools/locust-load-inference/sample-terraform.tfvars
```diff
@@ -9,7 +9,7 @@ ksa       = "benchmark-ksa"
 
 # Locust service configuration 
 artifact_registry                        = "us-central1-docker.pkg.dev/$PROJECT_ID/ai-benchmark"
-inference_server_service                 = "http://tgi" # inference server service name
+inference_server_service                 = "tgi" # inference server service name
```
hubatish, line 12: For my curiosity, how does this link work? Does kubernetes manage the mapping from service name -> ip address which this takes advantage of?

hubatish (approved):
Thanks for the maintenance & testing!

kfswain (approved):

PR #694, opened 2024-06-04
Support vllm openai api server

zmvictor:
Per https://docs.vllm.ai/en/latest/serving/metrics.html, openai api server supports vLLM serving metrics by default. This PR therefore:
- updates api server from vanilla to openai mode
- adds `swap_space` argument suggested in vLLM [benchmarks](https://github.com/vllm-project/vllm/blob/main/benchmarks/benchmark_serving.py#L6)

e2e tests with model `meta-llama/Llama-2-7b-chat-hf`. After [terraform apply](https://github.com/GoogleCloudPlatform/ai-on-gke/tree/main/benchmarks/inference-server/vllm#step-4-terraform-initialize-plan-and-apply):
```
# Get vLLM LB's external IP
$ VLLM_EXTERNAL_IP=`kubectl -n benchmark get service vllm -o jsonpath='{.status.loadBalancer.ingress[0].ip}'`

# send a prompt to the endpoint
$ curl $VLLM_EXTERNAL_IP/v1/completions \
    -H "Content-Type: application/json" \
    -d '{
        "model": "meta-llama/Llama-2-7b-chat-hf",
        "prompt": "Seattle City is a",
        "max_tokens": 7,
        "temperature": 0
    }'

# Check prometheus metrics
$ curl $VLLM_EXTERNAL_IP/metrics/

...
# TYPE vllm:prompt_tokens_total counter
vllm:prompt_tokens_total{model_name="meta-llama/Llama-2-7b-chat-hf"} 9.0
...

 
```

### DIFF
--- a/benchmarks/inference-server/vllm/main.tf
+++ b/benchmarks/inference-server/vllm/main.tf
@@ -51,6 +51,7 @@ resource "kubernetes_manifest" "default" {
     namespace                      = var.namespace
     model_id                       = var.model_id
     gpu_count                      = var.gpu_count
+    swap_space                     = var.swap_space
     ksa                            = var.ksa
     hugging_face_token_secret_list = local.hugging_face_token_secret == null ? [] : [local.hugging_face_token_secret]
   }))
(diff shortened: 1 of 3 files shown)


achandrasekar (approved):
Thanks for adding this!

annapendleton (approved):

achandrasekar:
Hi @zmvictor just remembered that this breaks the benchmark automation that we have for vLLM where we still are using the `/generate` API and not the `/completions` API - https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py#L172. It would be good to address that too.

PR #702, opened 2024-06-13
[TPU Provisioner] Add admission label selectors and e2e test script

nstogner:
* Avoid mutation when provisioning is disabled
* Avoid mutation for Jobs not managed by a JobSet
* Add e2e test script and cases


danielvegamyhre (approved):
LGTM! Thanks for doing this.

PR #705, opened 2024-06-17
Adding CSV output pushed by the Locust master

kfswain:


benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py
```diff
@@ -60,4 +66,17 @@ def json_dump_report(self):
             "average-time-to-first-token": sum(self.time_to_first_token_list)/max(len(self.time_to_first_token_list),1)
         }
         return json.dumps(stats)
+    
+    def dump_to_csv(self):
+        fields = ['success', 'total_request_time', 'time_to_first_token', 'input_tokens', 'output_tokens']
+        now = datetime.now()
+        bucket_name = "jetstream-benchmark-output-bucket-01"
```
achandrasekar, line 73: Can we use the bucket that the user already specified in the TF config here and just store it with a different file name?

kfswain: Agh, missed that, apologies. Done.

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
```diff
@@ -371,11 +376,11 @@ def intercept(
         # chunk sent back is used to calculate time to first token(TTFT).
         for response in responses:
             if ttft == 0:
-                ttft = time.perf_counter() - start_perf_counter
+                ttft = (time.perf_counter() - start_perf_counter) * 1000
```
annapendleton, line 379: Checking my own understanding of this :) Is this to convert ttft to ms? assuming this is a bug fix since we did the *1000 in the comments but not the actual var

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py
annapendleton: I still see hardcoded bucket, double checking commit was added in :)

annapendleton: Oh I see below - we should delete this line I think

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
```diff
@@ -339,13 +341,16 @@ def grpc_infer(self):
...
+        try:
+            output, ttft, response_time = self.stub.Decode(request)
+            logging.info(f"Response: {output}")
 
+            number_of_input_tokens = len(tokenizer.encode(prompt))
+            number_of_output_tokens = len(tokenizer.encode(output))
+            send_metrics(number_of_input_tokens, number_of_output_tokens, response_time, 1, ttft)
+        except:
+            # Capture that a test was ran, but the request threw an exception
+            send_metrics(-1,-1,-1,0,-1)
```
annapendleton, line 353: Wondering how doing -1 here will impact overall average data calculated in the MetricCollecter as it is averaged out over time. Do we discard the requests that throw an exception in those calculations?

annapendleton (changes requested):

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
kfswain: The metrics locust captures likely wouldnt be hit. As we would need to call `self.env.events.request.fire` for the request to be captures, and it's low probability that an exception would be thrown by the time it gets there. 

As far as scraping and averaging the metrics. That is true that it could miscalculate the averages. However, much of the same data is captured in the CSV, so these sentinel values could be filtered off. 

We may need to take another pass at the runner report. But I considered that out of scope.

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py
kfswain: Agreed, done.

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
kfswain: Yeah, the data is still technically accurate. But definitely inconsistent with the total response time, and a bit weird to reason about as a decimal

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
annapendleton: I was thinking more like we need to filter out the -1's in this calculation here: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/01630c62fac8e5230c7cc0f8ff338e8fb3da8a5a/benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py#L36

Reason being, these avg metrics are also continuously scraped to cloud monitoring for real time metrics, and has also been useful in the autoscaling benchmarking work outside of the csv work. If we include the -1 values in this calculation, that data will be tainted. So I think filtering out those metrics in this calculation is important to implement within this initial PR to ensure we don't introduce data errors into other streams of data collection.

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
annapendleton: sg ty for confirming!

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py
annapendleton: ty!

benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/tasks.py
kfswain: Ah, I get what you're saying now. This should be circumnavigated since I pass in a `0` for request successful: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/custom_metric_aggregator.py#L15

I capture all the request metrics in a dictionary to make sure all the metrics of one request stay linked.

annapendleton: Sweet, that lgtm!

annapendleton (approved):

PR #709, opened 2024-06-18
Module Deduplication and bash Equivalents

Bslabe123:
- Moved common terraform modules to a subfolder of the lowest common parent folder (./modules)
- Did the same with image source code and dockerfiles (./images). 
- Deduplicated checkpoint-jobs, provided a pattern to make these modules implementation agnostic (deployment files were YAML now JSON, decorated with `jq`).


tutorials-and-examples/inference-servers/jetstream/maxtext/single-host-inference/README.md
```diff
@@ -93,22 +93,22 @@ kubectl create secret generic kaggle-secret \
...
 3. Convert the checkpoint to a MaxText compatible checkpoint
 4. Unscan the checkpoint to be used for inference
 
 In the manifest, ensure the value of the BUCKET_NAME environment variable is the name of the Cloud Storage bucket you created above. Do not include the `gs://` prefix.
 
-Apply the manifest:
+Apply the manifest from the root of this repo to create the job:
 ```
-kubectl apply -f checkpoint-job.yaml
+BUCKET_NAME=<YOUR_BUCKET_NAME> IMPLEMENTATION="JAX" ./modules/checkpoint-converter/bash-equivalent/apply.sh
```
annapendleton, line 106: If you want to use bash scripts, you should migrate them to this jetstream maxtext single-host-inference directory.

A modules directory is a terraform standard naming schema indicating that the files within it are terraform scripts.

In this case, these bash scripts are specific tooling that works mostly for this tutorial rather than what is indicated to be reusable terraform modules.

images/http-server/Dockerfile
annapendleton, line 1: images is not a common directory in ai-on-gke yet. These images are specific to the tutorial with hardcoded configuration that works for the tutorial, rather than generic images that are applicable to other directories within ai-on-gke/ top level.

We should migrate these images back to the tutorials jetstream directory until another use case in a different directory comes up.

annapendleton (changes requested):

PR #713, opened 2024-06-25
Serve pylint source code on maxengine-server image to satisfy GPLv2 terms

Bslabe123:


tutorials-and-examples/inference-servers/jetstream/maxtext/maxengine-server/Dockerfile
```diff
@@ -17,6 +17,9 @@ RUN update-alternatives --install \
 
 RUN git clone https://github.com/google/maxtext.git
 
+# Pylint served on image to satisfy the terms of GPLv2.
+RUN git clone https://github.com/pylint-dev/pylint.git
```
vivianrwu, line 21: Perhaps we should not include this in the Dockerfile and just add it to the image, since this clone will add to Dockerfile building + deploying time and not relevant to the maxengine-server itself.

PR #715, opened 2024-06-27
ml-platform release from development branch

arueth:

PR #720, opened 2024-07-02
clean up TGI quickstart pathway

annapendleton:
Update readme and default vars for cleaning up quickstart README.md top level

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -88,7 +88,7 @@ cd inference-server/text-generation-inference
 # Copy the sample variables and update the project number and cluster name in
 # the fleet_host variable "https://connectgateway.googleapis.com/v1/projects/<project-number>/locations/global/gkeMemberships/<cluster-name>"
 # in the `terraform.tfvars` file.
-cp ./sample-tfvars/gpu-sample.tfvars terraform.tfvars
+cp ./sample-terraform.tfvars terraform.tfvars
 
 # Initialize the Terraform modules.
 terraform init
--- a/benchmarks/benchmark/tools/locust-load-inference/sample-tfvars/tgi-sample.tfvars
+++ b/benchmarks/benchmark/tools/locust-load-inference/sample-tfvars/tgi-sample.tfvars
@@ -2,16 +2,18 @@ credentials_config = {
   fleet_host = "https://connectgateway.googleapis.com/v1/projects/$PROJECT_NUM/locations/global/gkeMemberships/ai-benchmark"
 }
 
-project_id = "change-me"
+project_id = "$PROJECT_ID"
 
 namespace = "benchmark"
 ksa       = "benchmark-ksa"
+
+k8s_hf_secret = "hf-token"
 
 # Locust service configuration 
 artifact_registry                        = "us-central1-docker.pkg.dev/$PROJECT_ID/ai-benchmark"
 inference_server_service                 = "tgi" # inference server service name
 locust_runner_kubernetes_service_account = "sample-runner-ksa"
-output_bucket                            = "benchmark-output"
+output_bucket                            = "${PROJECT_ID}-benchmark-output"
 gcs_path                                 = "gs://${PROJECT_ID}-ai-gke-benchmark-fuse/ShareGPT_V3_unfiltered_cleaned_split_filtered_prompts.txt"
 
 # Benchmark configuration for Locust Docker accessing inference server
--- a/benchmarks/infra/stage-2/sample-tfvars/gpu-sample.tfvars
+++ b/benchmarks/infra/stage-2/sample-tfvars/gpu-sample.tfvars
@@ -6,7 +6,7 @@ credentials_config = {
 
 # can be obtained from stage-1 by running:
 # terraform output -json  | jq '."project_id".value'
-project_id = "change-me"
+project_id = "$PROJECT_ID"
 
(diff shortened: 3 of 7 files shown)


benchmarks/README.md
```diff
@@ -88,7 +88,7 @@ cd inference-server/text-generation-inference
 # Copy the sample variables and update the project number and cluster name in
 # the fleet_host variable "https://connectgateway.googleapis.com/v1/projects/<project-number>/locations/global/gkeMemberships/<cluster-name>"
 # in the `terraform.tfvars` file.
-cp ./sample-tfvars/gpu-sample.tfvars terraform.tfvars
+cp ./sample-terraform.tfvars terraform.tfvars
```
kfswain, line 91: I dont see a corresponding changing in pathing, odd merge perhaps?

benchmarks/benchmark/tools/locust-load-inference/sample-tfvars/tgi-sample.tfvars
```diff
@@ -2,17 +2,19 @@ credentials_config = {
...
 
+k8s_hf_secret = "hf-token"
+
 # Locust service configuration 
 artifact_registry                        = "us-central1-docker.pkg.dev/$PROJECT_ID/ai-benchmark"
 inference_server_service                 = "tgi" # inference server service name
 locust_runner_kubernetes_service_account = "sample-runner-ksa"
 output_bucket                            = "benchmark-output"
-gcs_path                                 = "gs://${PROJECT_ID}-ai-gke-benchmark-fuse/ShareGPT_V3_unfiltered_cleaned_split_filtered_prompts.txt"
+gcs_path                                 = "gs://ai-gke-benchmark-fuse/ShareGPT_V3_unfiltered_cleaned_split_filtered_prompts.txt"
```
kfswain, line 17: Did we make this a fully public bucket that anyone can use?

benchmarks/infra/stage-2/sample-tfvars/gpu-sample.tfvars
```diff
@@ -6,12 +6,12 @@ credentials_config = {
...
 # terraform output -json  | jq '."project_id".value'
-project_id = "change-me"
+project_id = "$PROJECT_ID"
 
-bucket_name     = "${PROJECT_ID}-ai-gke-benchmark-fuse"
+bucket_name     = "ai-gke-benchmark-fuse"
 bucket_location = "US"
 
-output_bucket_name     = "${PROJECT_ID}-benchmark-output"
+output_bucket_name     = "benchmark-output"
```
kfswain, line 14: I think this has potential naming collisions? I thought GCS buckets have to be globally unique

kfswain (approved):
LGTM for the most part, just minor questions

benchmarks/README.md
annapendleton: correct, this is a bug fix because tgi is a GPU only workload, so it doesn't have multiple tfvars

ref: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/benchmarks/inference-server/text-generation-inference/sample-terraform.tfvars

benchmarks/infra/stage-2/sample-tfvars/gpu-sample.tfvars
annapendleton: I think you're right, I gotta fix this before submitting - documentation here https://cloud.google.com/storage/docs/buckets#naming

My main goal for this change was it was easier to make 1 sample change vs many. I had a few steps where I missed a var to update :)

Wondering if we should prepend the project id in the module itself. Or possibly stack the variables that require a changed var next to each other at the top of the tfvars so it's more visible? WDYT?

benchmarks/benchmark/tools/locust-load-inference/sample-tfvars/tgi-sample.tfvars
annapendleton: This is not a public bucket, I'm not sure we want to host one yet. I addressed my goals in the comment below :) let's continue the conversation there

benchmarks/infra/stage-2/sample-tfvars/gpu-sample.tfvars
kfswain: I'm down to prepend on their behalf, we could just leave a comment on the section of vars we will do that for, so they are aware.

annapendleton: Sweet, I'm assuming it's something the jetstream example vars will also benefit from automatically, so I'll also update those configs when I get the prepending updated. Will ping when ready for review on that change :) thanks for catching this!

annapendleton: updated commit to have prepended project id. Plan is to  do a follow up PR when we have time for automating the bucket naming schema (either in the modules or possibly in the orchestration end to end workflow)

arueth:
Why was this pushed even though there was a failing check? This also breaks the checks for everyone else.

arueth:
I fixed the formatting issue in https://github.com/GoogleCloudPlatform/ai-on-gke/pull/715

annapendleton:
> I fixed the formatting issue in #715

Thanks @arueth! Apologies, our benchmarking directory was previously excluded from the tests and checks. Will make sure to pay closer attention next time :)

PR #722, opened 2024-07-04
Enable Ray Autoscaler for the Rag example application

gongmax:
Enable Ray Autoscaler for the Rag example application. Tested with both GKE Autopilot cluster and Standard cluster with Cluster Autoscaler.

### DIFF
--- a/applications/rag/variables.tf
+++ b/applications/rag/variables.tf
@@ -398,7 +398,7 @@ variable "gpu_pools" {
     name               = "gpu-pool-l4"
     machine_type       = "g2-standard-24"
     autoscaling        = true
-    min_count          = 1
+    min_count          = 0
     max_count          = 3
     disk_size_gb       = 200
     disk_type          = "pd-balanced"
--- a/modules/kuberay-cluster/values.yaml
+++ b/modules/kuberay-cluster/values.yaml
@@ -36,7 +36,7 @@ head:
   # If enableInTreeAutoscaling is true, the autoscaler sidecar will be added to the Ray head pod.
   # Ray autoscaler integration is supported only for Ray versions >= 1.11.0
   # Ray autoscaler integration is Beta with KubeRay >= 0.3.0 and Ray >= 2.0.0.
-  # enableInTreeAutoscaling: true
+  enableInTreeAutoscaling: true
   # autoscalerOptions is an OPTIONAL field specifying configuration overrides for the Ray autoscaler.
   # The example configuration shown below below represents the DEFAULT values.
   # autoscalerOptions:
@@ -95,17 +95,17 @@ head:
       # Ray recommends at least 8G memory for production workloads.
       memory: "8G"
       # Sum of ephemeral storage requests must be max 10Gi on Autopilot default class.
-      # This includes, ray-head, gcsfuse-sidecar, and fluent-bit.
-      ephemeral-storage: 4Gi
+      # This includes, ray-head, gcsfuse-sidecar, fluent-bit, and ray Autoscaler sidecar which requests 1Gi by default.
+      ephemeral-storage: 3Gi
     requests:
       cpu: "4"
       memory: "8G"
-      ephemeral-storage: 4Gi
+      ephemeral-storage: 3Gi
   annotations:
     gke-gcsfuse/volumes: "true"
     gke-gcsfuse/cpu-limit: "1"
     gke-gcsfuse/memory-limit: 2Gi
-    gke-gcsfuse/ephemeral-storage-limit: 4Gi
+    gke-gcsfuse/ephemeral-storage-limit: 3Gi
   nodeSelector:
     iam.gke.io/gke-metadata-server-enabled: "true"
   tolerations: []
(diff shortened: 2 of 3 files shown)


roberthbailey:
Both of the presubmit tests failed. Can you investigate and see if the failures are related to your PR?

modules/kuberay-cluster/values.yaml
```diff
@@ -95,17 +95,17 @@ head:
       # Ray recommends at least 8G memory for production workloads.
       memory: "8G"
       # Sum of ephemeral storage requests must be max 10Gi on Autopilot default class.
-      # This includes, ray-head, gcsfuse-sidecar, and fluent-bit.
-      ephemeral-storage: 4Gi
+      # This includes, ray-head, gcsfuse-sidecar, fluent-bit, and ray Autoscaler sidecar which requests 1Gi by default.
+      ephemeral-storage: 3Gi
```
imreddy13, line 99: why change the storage value?

modules/kuberay-cluster/values.yaml
```diff
@@ -95,17 +95,17 @@ head:
...
       cpu: "4"
       memory: "8G"
-      ephemeral-storage: 4Gi
+      ephemeral-storage: 3Gi
   annotations:
     gke-gcsfuse/volumes: "true"
     gke-gcsfuse/cpu-limit: "1"
     gke-gcsfuse/memory-limit: 2Gi
-    gke-gcsfuse/ephemeral-storage-limit: 4Gi
+    gke-gcsfuse/ephemeral-storage-limit: 3Gi
```
imreddy13, line 108: same?

modules/kuberay-cluster/values.yaml
```diff
@@ -36,7 +36,7 @@ head:
   # If enableInTreeAutoscaling is true, the autoscaler sidecar will be added to the Ray head pod.
   # Ray autoscaler integration is supported only for Ray versions >= 1.11.0
   # Ray autoscaler integration is Beta with KubeRay >= 0.3.0 and Ray >= 2.0.0.
-  # enableInTreeAutoscaling: true
+  enableInTreeAutoscaling: true
```
imreddy13, line 39: thanks for re-enabling it! Can we add a small section to the readme that details how to use the autoscaler? for eg, for a larger dataset, maybe a user would need more than the max allowed workers (5) and they should know how to change it.

imreddy13 (approved):

applications/rag/variables.tf
```diff
@@ -398,7 +398,7 @@ variable "gpu_pools" {
     name               = "gpu-pool-l4"
     machine_type       = "g2-standard-24"
     autoscaling        = true
-    min_count          = 1
+    min_count          = 0
```
imreddy13, line 401: should we have 1 min worker pre-created so the ray job isn't super slow?

modules/kuberay-cluster/values.yaml
gongmax: The Ray Autoscaler sidecar has 1Gi storage requests so we need to decrease the storage requests for some other containers so to stay with the total 10Gi ephemeral storage requests limit on Autopilot default class. I made the decrease for ray-head and gcsfuse simply because the number works and I also want to leave some extra room. I added an in-line comments to explain it. i.e. line 98.

applications/rag/variables.tf
gongmax: From my testings, it took around 5 mins to scale up one worker node from 0. I feel it's not very long and sounds like a reasonable trade for cost. But I don't have strong opinion.

modules/kuberay-cluster/values.yaml
gongmax: Before this PR, the replicas of the worker was set to 1 so the current change can probably cover our previous usage assumption. I'm a little hesitating to add more description around how to change these configuration because for the configurations in this file, we don't expose them to users (i.e. via workloads.tfvars) and we don't have a README of how/when to changing them, which gave me impression that we don't want or at least don't expect users to change them, since these configurations work with the example we provide. 
For you particular example, since right now we don't provide guidance on how to change the dataset, I think adding guidance on changing the configurations may not necessary.
But again, I don't have strong opinion. :)

gongmax: Or, we want to expose a variable in "application/rag/workloads.tfvars" and "application/ray/workloads.tfvars" to give users control to enable/disable the ray autoscaler? I feel that's a little over engineering for the purpose of the example application. And I didn't see much concern to just use ray autoscaler as the solution.

applications/rag/variables.tf
roberthbailey: We could also set the [initial node count](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#initial_node_count) to 1 and enable autoscaling from 0 -> 3 which would allow it to scale down to 0 when nothing was running but pre-create the node at the beginning when we know we are about to use it. 

That would be a bit more of an invasive change since we'd have to plumb a new variable into the template.

roberthbailey: Also, this change only applies when creating a standard cluster (which is no longer the default) since you don't create node pools on autopilot.

roberthbailey: Looking again, it looks like the initial_node_count defaults to 1 so that should already work nicely.

roberthbailey (approved):

applications/rag/variables.tf
gongmax: Yes, for standard cluster, it creates a gpu pool with 1 node default since initial_node_count is set.

PR #727, opened 2024-07-09
Deduplicate custom-metrics-stackdriver-adapter module

Bslabe123:


benchmarks/inference-server/text-generation-inference/custom-metrics-stackdriver-adapter/main.tf
```diff
@@ -1,278 +0,0 @@
...
-    name = "custom-metrics:system:auth-delegator"
-  }
-  role_ref {
-    api_group = "rbac.authorization.k8s.io"
-    kind      = "ClusterRole"
-    name      = "system:auth-delegator"
-  }
-  subject {
-    kind = "ServiceAccount"
-    name = (var.workload_identity.enabled
```
rsgowman, line 37: NB: The replacement (rolebinding_custom-metrics-auth-reader.yaml.tftpl) hard-codes the service account name. i.e. here: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/f2883eb600045e9614726c8bf28123dca15885d2/modules/custom-metrics-stackdriver-adapter/templates/rolebinding_custom-metrics-auth-reader.yaml.tftpl#L12

Technically, that means that terraform doesn't know that it should create the service account before the role binding. (Probably, terraform just executes blocks in order if there's no dependency, so I suspect it works anyways.)

To fix that, you could parameterize the `*.tftpl` files to enforce correct ordering.

This is unrelated to this PR though, as that ordering concern pre-exists.

rsgowman (approved):

PR #728, opened 2024-07-10
Simplify how default values are set for complex variables types.

roberthbailey:
As documented in https://developer.hashicorp.com/terraform/language/expressions/type-constraints#optional-object-type-attributes

"Terraform applies the default value you specify in the optional modifier first and then later applies any nested default values to that attribute"

which means that the values specified in "optional" constraints was **always** overridden immediately by the defaults specified below.

This change simplifies how the defaults are set, solely using the optional defaults.

### DIFF
--- a/applications/jupyter/variables.tf
+++ b/applications/jupyter/variables.tf
@@ -155,10 +155,10 @@ variable "autopilot_cluster" {
 
 variable "cpu_pools" {
   type = list(object({
-    name                   = string
-    machine_type           = string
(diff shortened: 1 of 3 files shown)


spencer-p:
@roberthbailey Is this PR still active? Can you resolve conflicts if so?

roberthbailey:
I had forgotten about this PR. At this point I'm going to close it and I will re-open a new one if this bothers me again in the future.

PR #733, opened 2024-07-12
Product Category Preprocessing added

karajendran:


best-practices/ml-platform/examples/use-case/ray/dataprocessing/src/preprocessing.py
```diff
@@ -1,35 +1,36 @@
...
-import sys
 import time
 import urllib.error
 import urllib.request
+import numpy as np
 
 from google.cloud import storage
 from google.cloud.storage.retry import DEFAULT_RETRY
 from typing import List
-
```
arueth, line 17: Can you leave this blank line in to delineate between the imports and variables?

arueth (commented):
We should probably choose a coding style and a formatter/linter for the code to standardize on.

arueth (approved):

PR #723, opened 2024-07-04
Ray TPU Webhook Reliability Improvements

ryanaoleary:
This PR depends on changes from https://github.com/GoogleCloudPlatform/ai-on-gke/pull/740 which should be merged first ([c4cdf31](https://github.com/GoogleCloudPlatform/ai-on-gke/pull/723/commits/c4cdf319e890eafe0c303564214a757b9359e056) marks the start of the changes in this PR).

This PR improves the reliability of the webhook by making it stateless in between calls, fixing issues related to the `sliceToWorkers` mapping being cleared upon webhook restart. These changes rely on adding a [k8s PodInformer](https://pkg.go.dev/k8s.io/client-go@v0.30.2/informers/core/v1#PodInformer) to the webhook that watches Pods in the GKE cluster with the `ray.io/node-type=worker` label. We can then determine the next `replicaIndex` and `TPU_WORKER_ID` using the PodInformer cache. These changes remove the need to intercept Pod deletion requests.

This PR has been tested as follows:
- [x] Unit Tests
- [x] Manual Tests using [single-host](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.tpu-v4-singlehost.yaml), [multi-host](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.tpu-v4-multihost.yaml), multi-slice, and an [autoscaling RayCluster](https://github.com/ray-project/kuberay/blob/v1.1.1/ray-operator/config/samples/ray-cluster.autoscaler.yaml) with a TPU worker group added, intercepting additional Pods after webhook restart, and multiple webhook replicas with a Kuberay operator restart

PR #740, opened 2024-07-15
Ray TPU Webhook Autoscaling Support

ryanaoleary:
This PR adds support for autoscaling RayClusters by generating `TPU_WORKER_HOSTNAMES` when intercepting each Pod, rather than when intercepting the RayCluster CR.

This PR has been tested as follows:

- [x] Unit Tests
 - [x] Manual Tests using [single-host](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.tpu-v4-singlehost.yaml), [multi-host](https://github.com/ray-project/kuberay/blob/master/ray-operator/config/samples/ray-cluster.tpu-v4-multihost.yaml), and an [autoscaling RayCluster](https://github.com/ray-project/kuberay/blob/v1.1.1/ray-operator/config/samples/ray-cluster.autoscaler.yaml) with a TPU worker group added

### DIFF
--- a/ray-on-gke/tpu/kuberay-tpu-webhook/main.go
+++ b/ray-on-gke/tpu/kuberay-tpu-webhook/main.go
@@ -50,9 +50,6 @@ var (
 	// map of pod slices to workers in the slice
 	sliceToWorkers map[slice][]worker
 
-	// map of pod slices to TPU_WORKER_HOSTNAMES in that pod slice
-	sliceToHostnames map[slice]string
-
 	// Flag arguments.
 	BindAddr   string
 	CACert     string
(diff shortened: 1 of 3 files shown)


ray-on-gke/tpu/kuberay-tpu-webhook/main.go
```diff
@@ -134,18 +131,18 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
...
 }
 
-func genDNSHostnames(workerGroupSpec ray.WorkerGroupSpec, clusterName string, namespace string, replicaIndex int) (string, error) {
-	numHosts := workerGroupSpec.NumOfHosts
-	if numHosts == 0 {
-		return "", errors.New("workerGroupSpec NumOfHosts not set")
+func genDNSHostnames(numOfHosts int32, groupName string, clusterName string, namespace string, replicaIndex int) (string, error) {
+	if numOfHosts == 0 {
+		err := errors.New("workerGroupSpec NumOfHosts not set")
+		klog.ErrorS(err, "genDNSHostnames", "RayCluster", namespace+"/"+clusterName, "NumOfHosts", numOfHosts)
```
andrewsykim, line 137: just return an error instead

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
```diff
@@ -134,18 +131,18 @@ func extractRayCluster(admissionReview *admissionv1.AdmissionReview) (*ray.RayCl
 	return &rayCluster, nil
 }
 
-func genDNSHostnames(workerGroupSpec ray.WorkerGroupSpec, clusterName string, namespace string, replicaIndex int) (string, error) {
-	numHosts := workerGroupSpec.NumOfHosts
-	if numHosts == 0 {
-		return "", errors.New("workerGroupSpec NumOfHosts not set")
+func genDNSHostnames(numOfHosts int32, groupName string, clusterName string, namespace string, replicaIndex int) (string, error) {
+	if numOfHosts == 0 {
```
andrewsykim, line 135: numOfHosts = 0 is still valid with TPUs though right?

andrewsykim:
> This PR adds support for autoscaling RayClusters by generating TPU_WORKER_HOSTNAMES when intercepting each Pod, rather than when intercepting the RayCluster CR.

Will existing pods need to have the `TPU_WORKER_HOSTNAMES` env var updated when new replicas are added or does it only apply for new pods added by the autoscaler?

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
ryanaoleary: I don't think `numOfHosts = 0` is ever valid, `numOfHosts` defaults to 1 and if a user sets it to 0 then no workers would be created per replica.

ryanaoleary:
> > This PR adds support for autoscaling RayClusters by generating TPU_WORKER_HOSTNAMES when intercepting each Pod, rather than when intercepting the RayCluster CR.
> 
> Will existing pods need to have the `TPU_WORKER_HOSTNAMES` env var updated when new replicas are added or does it only apply for new pods added by the autoscaler?

We don't need to update existing Pods. We assume that TPU PodSlices are scaled atomically so the `TPU_WORKER_HOSTNAMES` we add to each Pod include the DNS hostnames for all the Pods in the slice, even if they haven't been created yet.

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
ryanaoleary: Done in [27ddd5d](https://github.com/GoogleCloudPlatform/ai-on-gke/pull/740/commits/27ddd5db325ab7a3856f5b8223e85d9fcb16295f).

andrewsykim (approved):

andrewsykim:
(let's ensure https://github.com/GoogleCloudPlatform/ai-on-gke/pull/723 is also merged before cutting a new tag with the autoscaling changes)

PR #746, opened 2024-07-19
Add Kueue exemplary setup for reservation and DWS

PBundyra:
Added Kueue exemplary setup which uses reservation nodepool first, and if there is no capacity in reservation, then tries DWS.


alculquicondor (commented):
/lgtm
/approve

alculquicondor (approved):

andrewsykim:
The cloud build presubmit can be ignored for this PR since we don't test the DWS example there

PR #757, opened 2024-07-30
Add quantization support for TGI

achandrasekar:
This change adds quantization support so we can use quantization techniques like eetq, bitsandbytes, etc. when deploying models using TGI. It allows us to benchmark models like Llama3 405B using FP8 too.

### DIFF
--- a/benchmarks/inference-server/text-generation-inference/variables.tf
+++ b/benchmarks/inference-server/text-generation-inference/variables.tf
@@ -70,6 +70,16 @@ variable "max_concurrent_requests" {
   }
 }
 
+variable "quantization" {
+  description = "Quantization used for the model."
+  type        = string
+  nullable    = true
+  # Can be one of the quantization options mentioned in https://huggingface.co/docs/text-generation-inference/en/basic_tutorials/launcher#quantize.
+  # `eetq` and `bitsandbytes` can be applied to any models whereas others might
+  # require the use of quantized checkpoints.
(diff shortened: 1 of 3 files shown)


benchmarks/inference-server/text-generation-inference/variables.tf
```diff
@@ -70,6 +70,16 @@ variable "max_concurrent_requests" {
   }
 }
 
+variable "quantization" {
+  description = "Quantization used for the model."
+  type        = string
+  nullable    = true
+  # Can be one of the quantization options mentioned in https://huggingface.co/docs/text-generation-inference/en/basic_tutorials/launcher#quantize.
```
annapendleton, line 77: small q, wondering if this is useful to add to the description itself?

we could also consider putting in validation as well with the options, ex: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/fdef210fbdfcb64e00e9fa17f6f346c2c3265ad6/benchmarks/inference-server/text-generation-inference/variables.tf#L55

annapendleton (approved):

benchmarks/inference-server/text-generation-inference/variables.tf
achandrasekar: Yes, moving it into description makes sense. Updated. I didn't want to add validation since it requires us to keep it updated with any changes TGI makes.

PR #761, opened 2024-07-31
TPU Metrics PodMonitoring

Bslabe123:
Added PodMonitoring to scrape TPU metrics with higher frequency (15s)

### DIFF
--- a/modules/jetstream-maxtext-deployment/README.md
+++ b/modules/jetstream-maxtext-deployment/README.md
@@ -58,22 +58,18 @@ cat ./templates/podmonitoring.yaml.tftpl >> "$PODMONITORING_MANIFEST"
 PODMONITORING_TPU_MANIFEST=$(mktemp)
 cat ./templates/podmonitoring-tpu.yaml.tftpl >> "$PODMONITORING_TPU_MANIFEST"
 
-if [ "$METRICS_PORT" != "" ] && ([ "$SERVER_METRICS_SCRAPE_INTERVAL" != "" ] || [ "$SYSTEM_METRICS_SCRAPE_INTERVAL" != "" ]); then
-    if [ "$SERVER_METRICS_SCRAPE_INTERVAL" != "" ]; then
-        cat $PODMONITORING_MANIFEST \
-        | sed "s/\${metrics_port}/$METRICS_PORT/g" \
-        | sed "s/\${metrics_scrape_interval}/$SERVER_METRICS_SCRAPE_INTERVAL/g" >> "$PODMONITORING_MANIFEST"
-    fi
-
-    if [ "$SYSTEM_METRICS_SCRAPE_INTERVAL" != "" ]; then
-        cat $PODMONITORING_TPU_MANIFEST \
-        | sed "s/\${metrics_scrape_interval}/$SYSTEM_METRICS_SCRAPE_INTERVAL/g" >> "$PODMONITORING_TPU_MANIFEST"
-    fi
-
-    cat $JETSTREAM_MANIFEST | sed "s/\${metrics_port_arg}/prometheus_port=$METRICS_PORT/g" >> "$JETSTREAM_MANIFEST"
-    
-    cat $PODMONITORING_MANIFEST | kubectl apply -f -
+if [ "$SYSTEM_METRICS_SCRAPE_INTERVAL" != "" ]; then
+    cat $PODMONITORING_TPU_MANIFEST \
+    | sed "s/\${metrics_scrape_interval}/$SYSTEM_METRICS_SCRAPE_INTERVAL/g" >> "$PODMONITORING_TPU_MANIFEST"
     cat $PODMONITORING_TPU_MANIFEST | kubectl apply -f -
+fi
+
+if [ "$METRICS_PORT" != "" ] && [ "$SERVER_METRICS_SCRAPE_INTERVAL" != "" ]; then
+    cat $PODMONITORING_MANIFEST \
+    | sed "s/\${metrics_port}/$METRICS_PORT/g" \
+    | sed "s/\${metrics_scrape_interval}/$SERVER_METRICS_SCRAPE_INTERVAL/g" >> "$PODMONITORING_MANIFEST"
+    cat $PODMONITORING_MANIFEST | kubectl apply -f -
+    cat $JETSTREAM_MANIFEST | sed "s/\${metrics_port_arg}/prometheus_port=$METRICS_PORT/g" >> "$JETSTREAM_MANIFEST"
 else
     cat $JETSTREAM_MANIFEST | sed "s/\${metrics_port_arg}//g" >> "$JETSTREAM_MANIFEST"
 fi
--- a/modules/jetstream-maxtext-deployment/README.md
+++ b/modules/jetstream-maxtext-deployment/README.md
@@ -6,7 +6,8 @@ Assure the following environment variables are set:
    - MODEL_NAME: The name of your LLM (as of the writing of this README valid options are "gemma-7b", "llama2-7b", "llama2-13b")
    - PARAMETERS_PATH: Where to find the parameters for your LLM (if using the checkpoint-converter it will be "gs:\/\/$BUCKET_NAME\/final\/unscanned\/gemma_7b-it\/0\/checkpoints\/0\/items" where $BUCKET_NAME is the same one used in the checkpoint-converter)
    - (optional) METRICS_PORT: Port to emit custom metrics on
-   - (optional) METRICS_SCRAPE_INTERVAL: How often to scrape metrics
+   - (optional) SERVER_METRICS_SCRAPE_INTERVAL: How often to scrape Jetstream server metrics
+   - (optional) SYSTEM_METRICS_SCRAPE_INTERVAL: How often to scrape TPU system metrics
    - (optional) TPU_TOPOLOGY: Topology of TPU chips used by jetstream (default: "2x4")
    - (optional) TPU_TYPE: Type of TPUs used (default: "tpu-v5-lite-podslice")
    - (optional) TPU_CHIP_COUNT: Number of TPU chips requested, can be obtained by algebraically evaluating TPU_TOPOLOGY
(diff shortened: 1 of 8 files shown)


modules/jetstream-maxtext-deployment/main.tf
```diff
@@ -43,13 +44,20 @@ resource "kubernetes_manifest" "jetstream-service" {
 }
 
 resource "kubernetes_manifest" "jetstream-podmonitoring" {
-  count = var.maxengine_deployment_settings.metrics_port != null ? 1 : 0
+  count = var.maxengine_deployment_settings.metrics_port != null && var.maxengine_deployment_settings.metrics_scrape_interval != null ? 1 : 0
```
annapendleton, line 47: Why are we using the same metrics scrape interval for the jetstream workload metrics and the tpu accelerator metrics?

annapendleton, line 47: This && behavior mixed with the metrics_scrape_interval not having a default set means that this variable is required to be set to deploy the jetstream workload podmonitoring object.

Is this the desired behavior, or should there be a default metrics scrape interval so this will deploy when a metrics port is set?

modules/jetstream-maxtext-deployment/main.tf
```diff
@@ -43,13 +44,20 @@ resource "kubernetes_manifest" "jetstream-service" {
...
   manifest = yamldecode(templatefile(local.podmonitoring_template, {
     metrics_port            = var.maxengine_deployment_settings.metrics_port != null ? var.maxengine_deployment_settings.metrics_port : "",
     metrics_scrape_interval = var.maxengine_deployment_settings.metrics_scrape_interval
   }))
 }
 
+resource "kubernetes_manifest" "jetstream-podmonitoring-tpu" {
+  count = var.maxengine_deployment_settings.metrics_scrape_interval != null ? 1 : 0
+  manifest = yamldecode(templatefile(local.podmonitoring_tpu_template, {
+    metrics_scrape_interval = max(15, var.maxengine_deployment_settings.metrics_scrape_interval)
```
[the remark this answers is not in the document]
Bslabe123: Wanted to keep the scrape intervals between the two similar but two things, [GMP in general cant scape more frequently than 5s](https://cloud.google.com/stackdriver/docs/managed-prometheus#:~:text=Managed%20Service%20for%20Prometheus%20supports%20a%20minimum%20scrape%20interval%20of%205%20seconds.), and TPU metrics are only updated every 15 seconds

modules/jetstream-maxtext-deployment/templates/deployment.yaml.tftpl
```diff
@@ -38,6 +38,7 @@ spec:
         - load_parameters_path=${load_parameters_path_arg}
         - ${metrics_port_arg}
         ports:
+        - containerPort: 8431 # Port to export TPU runtime metrics, if supported.
```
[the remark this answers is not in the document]
Bslabe123: [I originally thought so too but 8431 is a hardcoded port for TPU mertics](https://cloud.google.com/kubernetes-engine/docs/how-to/tpus-autopilot#:~:text=In%20GKE%20version%201.27.4%2Dgke.900%20or%20later%2C%20TPU%20workloads%20that%20use%20JAX%20version%200.4.14%20or%20later%20and%20specify%20containerPort%3A%208431%20export%20TPU%20utilization%20metrics%20as%20GKE%20system%20metrics.%20The%20following%20metrics%20are%20available%20in%20Cloud%20Monitoring%20to%20monitor%20your%20TPU%20workload%27s%20runtime%20performance%3A)

modules/jetstream-maxtext-deployment/main.tf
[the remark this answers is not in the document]
Bslabe123: Good catch on the default! Removed that. Also the `var.maxengine_deployment_settings.metrics_scrape_interval` needs to be defined or count gets set to 0, plus we rely on the validator to make sure its an acceptable input

modules/jetstream-maxtext-deployment/main.tf
Bslabe123: Didn't see a reason to set them to different values (other than the 15s minimum for one), thoughts on the approach?

Bslabe123: The idea is that if the user doesn't want to scrape metics at all they leave the scrape interval field empty, but good call we should make the comments more descriptive. Likely going to follow this PR up at some point by reorganizing this modules parameters

annapendleton: Main reason that comes to my mind would be cost - if a user only wants to autoscale on one of the metrics, they may want to minimize costs by choosing a less aggressive time sampling for other metrics. See the pricing page here for the difference it can make: https://cloud.google.com/stackdriver/docs/managed-prometheus/cost-controls#scraping-interval

I agree that the different minimum requirements is an additional reason to separate them.

You can decide what you want to do here, I just wanted to understand why you wanted to keep it in one vs two :)

modules/jetstream-maxtext-deployment/main.tf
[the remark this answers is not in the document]
annapendleton: Ah good point, and makes sense. You answered my question in another question that the metrics_scrape_interval is required for the pod monitoring specs to be deployed :)

modules/jetstream-maxtext-deployment/variables.tf
```diff
@@ -31,8 +31,8 @@ variable "maxengine_deployment_settings" {
 
     model_name              = string           // Name of your LLM (for example: "gemma-7b")
     parameters_path         = string           // Path to the paramters for your model
-    metrics_port            = optional(number) // Emit Jetstream metrics on this port of each container
-    metrics_scrape_interval = optional(number) // Interval for scraping metrics (default: 10s)
+    metrics_port            = optional(number) // Emit Jetstream server metrics on this port of each container, no server metrics will be emitted if not set
```
annapendleton, line 34: will you add these additional details to the README description as well?

modules/jetstream-maxtext-deployment/main.tf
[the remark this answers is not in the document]
annapendleton: Ah I see - can you update in the comments about this 15s minimum for the tpu server metrics?

modules/jetstream-maxtext-deployment/main.tf
annapendleton: sgtm, comment updates sounds like a good enough solution for now to clarify the usage :)

modules/jetstream-maxtext-deployment/templates/deployment.yaml.tftpl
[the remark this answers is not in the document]
annapendleton: Ah, so the metrics port arg is only relevant for the tpu server metrics, ty for clarifying that in the comments!

liurupeng:
Nice! thanks for the reviewing! @annapendleton

modules/jetstream-maxtext-deployment/variables.tf
Bslabe123: Added

modules/jetstream-maxtext-deployment/main.tf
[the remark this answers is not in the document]
Bslabe123: Restructured metrics config in the module, guardrails are more specific and descriptive now

modules/jetstream-maxtext-deployment/README.md
```diff
@@ -53,11 +55,21 @@ cat ./templates/deployment.yaml.tftpl >> "$JETSTREAM_MANIFEST"
 PODMONITORING_MANIFEST=$(mktemp)
 cat ./templates/podmonitoring.yaml.tftpl >> "$PODMONITORING_MANIFEST"
 
-if [ "$METRICS_PORT" != "" ]; then
-    cat $PODMONITORING_MANIFEST | sed "s/\${metrics_port}/$METRICS_PORT/g" >> "$PODMONITORING_MANIFEST"
+PODMONITORING_TPU_MANIFEST=$(mktemp)
+cat ./templates/podmonitoring-tpu.yaml.tftpl >> "$PODMONITORING_TPU_MANIFEST"
+
+if [ "$METRICS_PORT" != "" ] && [ "$METRICS_SCRAPE_INTERVAL" != "" ]; then
```
annapendleton, line 61: Should this if statement be updated since METRICS_SCRAPE_INTERVAL is no longer defined?

modules/jetstream-maxtext-deployment/variables.tf
```diff
@@ -29,10 +29,18 @@ variable "maxengine_deployment_settings" {
     maxengine_server_image      = optional(string, "us-docker.pkg.dev/cloud-tpu-images/inference/maxengine-server:v0.2.2")
     jetstream_http_server_image = optional(string, "us-docker.pkg.dev/cloud-tpu-images/inference/jetstream-http:v0.2.2")
 
-    model_name              = string           // Name of your LLM (for example: "gemma-7b")
-    parameters_path         = string           // Path to the paramters for your model
-    metrics_port            = optional(number) // Emit Jetstream metrics on this port of each container
-    metrics_scrape_interval = optional(number) // Interval for scraping metrics (default: 10s)
+    model_name      = string // Name of your LLM (for example: "gemma-7b")
+    parameters_path = string // Path to the paramters for your model
```
annapendleton, line 33: nit: paramters -> parameters

modules/jetstream-maxtext-deployment/variables.tf
```diff
@@ -47,8 +55,13 @@ variable "maxengine_deployment_settings" {
...
   validation {
-    condition     = !(var.maxengine_deployment_settings.metrics_scrape_interval < 5)
-    error_message = "Metrics scrape interval must be at least 5s"
+    condition     = try(var.maxengine_deployment_settings.metrics.server.scrape_interval >= 5, true)
+    error_message = "Server metrics scrape interval may not be shorter than 5s"
+  }
+
+  validation {
+    condition     = try(var.maxengine_deployment_settings.metrics.system.scrape_interval >= 15, true)
+    error_message = "TPU system metrics scrape interal may not be shorter than 15s"
```
annapendleton, line 64: nit: interal -> interval

annapendleton (approved):

PR #769, opened 2024-08-06
Refactor: move workload identity service account out of kuberay-operator

genlu2011:
Move provisioning service accounts away from kuberay-operator module. These two are different tasks and after we use ray addon, we will not install kuberay-operator module in applications.

### DIFF
--- a/applications/rag/main.tf
+++ b/applications/rag/main.tf
@@ -245,7 +245,8 @@ module "kuberay-monitoring" {
   create_namespace                = true
   enable_grafana_on_ray_dashboard = var.enable_grafana_on_ray_dashboard
   k8s_service_account             = local.ray_service_account
-  depends_on                      = [module.namespace, module.kuberay-operator, module.kuberay-workload-identity]
+  //TODO(genlu): remove the module.kuberay-operator after migrated using ray addon.
+  depends_on = [module.namespace, module.kuberay-operator, module.kuberay-workload-identity]
 }
 
 module "kuberay-cluster" {
@@ -280,7 +281,7 @@ module "kuberay-cluster" {
   k8s_backend_service_port = var.ray_dashboard_k8s_backend_service_port
   domain                   = var.ray_dashboard_domain
   members_allowlist        = var.ray_dashboard_members_allowlist != "" ? split(",", var.ray_dashboard_members_allowlist) : []
-
+  //TODO(genlu): remove the module.kuberay-operator after migrated using ray addon.
   depends_on = [module.gcs, module.kuberay-operator, module.kuberay-workload-identity]
 }
 
--- a/applications/rag/main.tf
+++ b/applications/rag/main.tf
@@ -153,15 +153,13 @@ module "namespace" {
 }
 
 module "kuberay-operator" {
-  source                 = "../../modules/kuberay-operator"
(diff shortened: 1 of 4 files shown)


applications/rag/main.tf
```diff
@@ -225,6 +223,31 @@ module "kuberay-logging" {
...
+module "kuberay-workload-identity" {
+  providers                       = { kubernetes = kubernetes.rag }
+  source                          = "terraform-google-modules/kubernetes-engine/google//modules/workload-identity"
+  version                         = "30.0.0" # Pinning to a previous version as current version (30.1.0) showed inconsitent behaviour with workload identity service accounts
+  use_existing_gcp_sa             = !var.create_ray_service_account
+  name                            = local.ray_service_account
+  namespace                       = local.kubernetes_namespace
+  project_id                      = var.project_id
+  roles                           = ["roles/cloudsql.client", "roles/monitoring.viewer"]
+  automount_service_account_token = true
```
imreddy13, line 235: is this needed? I don't think we have it for jupyterhub or the frontend

applications/rag/main.tf
```diff
@@ -225,6 +223,31 @@ module "kuberay-logging" {
...
+module "kuberay-monitoring" {
+  source                          = "../../modules/kuberay-monitoring"
+  providers                       = { helm = helm.rag, kubernetes = kubernetes.rag }
+  project_id                      = var.project_id
+  autopilot_cluster               = local.enable_autopilot
+  namespace                       = local.kubernetes_namespace
+  create_namespace                = true
+  enable_grafana_on_ray_dashboard = var.enable_grafana_on_ray_dashboard
+  k8s_service_account             = local.ray_service_account
+  depends_on                      = [module.namespace, module.kuberay-operator, module.kuberay-workload-identity]
```
imreddy13, line 248: do we still need the dependency on kuberay operator here? or will we remove this once we switch to the addon?

applications/rag/main.tf
```diff
@@ -256,19 +280,8 @@ module "kuberay-cluster" {
...
-  providers                       = { helm = helm.rag, kubernetes = kubernetes.rag }
-  project_id                      = var.project_id
-  autopilot_cluster               = local.enable_autopilot
-  namespace                       = local.kubernetes_namespace
-  create_namespace                = true
-  enable_grafana_on_ray_dashboard = var.enable_grafana_on_ray_dashboard
-  k8s_service_account             = local.ray_service_account
-  # TODO(umeshkumhar): remove kuberay-operator depends, figure out service account dependency
-  depends_on = [module.namespace, module.kuberay-operator]
+  depends_on = [module.gcs, module.kuberay-operator, module.kuberay-workload-identity]
```
imreddy13, line 284: same question, if yes let's leave a TODO or create a bug to go do this later

applications/rag/main.tf
genlu2011: we will remove this when we switch to the addon

applications/rag/main.tf
genlu2011: not needed, removed.

applications/rag/main.tf
genlu2011: added todo

imreddy13 (approved):

PR #772, opened 2024-08-08
Decrease scrape interval of metrics from TGI and DCGM to 15s in benchmark

raywainman:
Decrease scrape interval of metrics from TGI and DCGM from 30s to 15s in benchmark, this will allow for quicker autoscaling and also for better graphs when plotting the data.

To do this, we need to export data out of DCGM more often (here I have it set to 5s by setting the [`collect-interval`](https://github.com/NVIDIA/dcgm-exporter/blob/6d499c68a93677e8ccafe857427b09d4617ebfae/pkg/cmd/app.go#L56) flag).

### DIFF
--- a/benchmarks/inference-server/text-generation-inference/hpa-templates/dcgm-podmonitoring.yaml.tftpl
+++ b/benchmarks/inference-server/text-generation-inference/hpa-templates/dcgm-podmonitoring.yaml.tftpl
@@ -26,7 +26,7 @@ spec:
       app: nvidia-dcgm-exporter
   endpoints:
     - port: metrics
-      interval: 30s
+      interval: 15s
       metricRelabeling:
         # Change the DCGM metric name that we want to use in HPA to lowercase.
         # This is because HPA doesn't work with uppercase external metrics:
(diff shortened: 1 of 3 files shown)


benchmarks/inference-server/text-generation-inference/hpa-templates/dcgm-podmonitoring.yaml.tftpl
```diff
@@ -26,7 +26,7 @@ spec:
       app: nvidia-dcgm-exporter
   endpoints:
     - port: metrics
-      interval: 30s
+      interval: 15s
```
annapendleton, line 29: not a blocker for this PR, wondering if we should turn this into a variable as well to make it easy to play with for future experiments.

annapendleton (approved):

achandrasekar (approved):

benchmarks/inference-server/text-generation-inference/hpa-templates/dcgm-podmonitoring.yaml.tftpl
raywainman: Filed #783.

PR #775, opened 2024-08-12
[WIP] Add latency profile generator

achandrasekar:
This change adds a new benchmarking suite called latency-profile-generator which runs serving benchmarks at different request rates to produce latency and throughput numbers at different QPS. This can be used to identify how different models and model servers perform depending on incoming traffic.

### DIFF
--- a/benchmarks/benchmark/tools/latency-profile/README.md
+++ b/benchmarks/benchmark/tools/latency-profile/README.md
@@ -1,0 +1,209 @@
+# AI on GKE Benchmark Locust
+
+<!-- BEGIN TOC -->
+- [AI on GKE Benchmark Locust](#ai-on-gke-benchmark-locust)
+  - [Overview](#overview)
+  - [Instructions](#instructions)
+    - [Step 1: prepare benchmark prompts](#step-1-prepare-benchmark-prompts)
+    - [Step 2: create and give service account access to view dataset](#step-2-create-and-give-service-account-access-to-view-dataset)
+    - [Step 3: create output bucket](#step-3-create-output-bucket)
+    - [Step 4: create and give service account access to write to output gcs bucket](#step-4-create-and-give-service-account-access-to-write-to-output-gcs-bucket)
+    - [Step 5: create artifact repository for automated Locust docker build](#step-5-create-artifact-repository-for-automated-locust-docker-build)
+    - [Step 6: create and configure terraform.tfvars](#step-6-create-and-configure-terraformtfvars)
+      - [\[optional\] set-up credentials config with kubeconfig](#optional-set-up-credentials-config-with-kubeconfig)
+      - [\[optional\] set up secret token in Secret Manager](#optional-set-up-secret-token-in-secret-manager)
+    - [Step 7: login to gcloud](#step-7-login-to-gcloud)
+    - [Step 8: terraform initialize, plan and apply](#step-8-terraform-initialize-plan-and-apply)
+    - [Step 9: start an end to end benchmark](#step-9-start-an-end-to-end-benchmark)
+      - [option 1: initiate a single end to end Locust benchmark run via curl command](#option-1-initiate-a-single-end-to-end-locust-benchmark-run-via-curl-command)
+      - [option 2: interactive benchmark with locust web ui](#option-2-interactive-benchmark-with-locust-web-ui)
+    - [Step 10: viewing metrics](#step-10-viewing-metrics)
+    - [Additional Tips](#additional-tips)
+  - [Variables](#variables)
+<!-- END TOC -->
+
+## Overview
+
+This deploys the latency profile generator which measures the throuhghput and
+latency at various request rates for the model and model server of your choice. 
+
+It currently supports the following frameworks:
+- tensorrt_llm_triton
+- text generation inference (tgi)
+- vllm
+- sax
+
+## Instructions
+
+### Step 1: create output bucket
+
+If you followed steps from `../../infra/` for creating your cluster and extra
+resources, you will already have an output bucket created for you.
+If not, you will have to create and manage your own gcs bucket for storing
+benchmarking results.
+
+Set the `output_bucket` in your `terraform.tfvars` to this gcs bucket.
+
+### Step 2: create and give service account access to write to output gcs bucket
+
+The Latency profile generator requires storage.admin access to write output to
+the given output gcs bucket. If you followed steps in `../../infra`, then you
+already have a kubernetes and gcloud service account created that has the proper
+access to the created output bucket.
+
+To give viewer permissions on the gcs bucket to the gcloud service account,
+run the following:
+
+```
+gcloud storage buckets add-iam-policy-binding  gs://$OUTPUT_BUCKET/
+--member=serviceAccount:$GOOGLE_SERVICE_ACCOUNT@$PROJECT_ID.iam.gserviceaccount.com --role=roles/storage.admin
+```
+
+Your kubernetes service account will inherit the reader permissions.
+
+You will set the `lantency_profile_kubernetes_service_account` in your
+`terraform.tfvars` to the kubernetes service account name.
+
+### Step 5: create artifact repository for automated Locust docker build
+
+The latency profile generator rebuilds the docker file on each terraform apply.
+The containers will be pushed to the given `artifact_registry`. This artifact
+repository is expected to already exist. If you created your cluster via
+`../../infra/`, then an artifact repository was created for you with the same
+name as the prefix in the same location as the cluster. You can also create your
+own via this command:
+
+```bash
+gcloud artifacts repositories create ai-benchmark --location=us-central1 --repository-format=docker
+```
+
+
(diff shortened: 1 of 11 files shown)


benchmarks/benchmark/tools/latency-profile/README.md
```diff
@@ -0,0 +1,199 @@
...
+gcloud storage buckets add-iam-policy-binding  gs://$OUTPUT_BUCKET/
+--member=serviceAccount:$GOOGLE_SERVICE_ACCOUNT@$PROJECT_ID.iam.gserviceaccount.com --role=roles/storage.admin
+```
+
+Your kubernetes service account will inherit the reader permissions.
+
+You will set the `lantency_profile_kubernetes_service_account` in your
+`terraform.tfvars` to the kubernetes service account name.
+
+### Step 5: create artifact repository for automated Latency Profile Generator docker build
```
Bslabe123, line 60: Missing steps 3 and 4?

benchmarks/benchmark/tools/latency-profile/manifest-templates/latency-profile-generator.yaml.tpl
```diff
@@ -0,0 +1,43 @@
...
+              value: ${inference_server_service}
+            - name: BACKEND
+              value: ${inference_server_framework}
+            - name: INPUT_LENGTH
+              value: ${max_prompt_len}
+            - name: OUTPUT_LENGTH
+              value: ${max_output_len}
+            - name: OUTPUT_BUCKET
+              value: ${output_bucket}
+%{ for hugging_face_token_secret in hugging_face_token_secret_list ~}
```
annapendleton, line 37: Since you have support for k8s_hf_secret, you'll need to add something like this code here to support it in this spec:

https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/benchmarks/benchmark/tools/locust-load-inference/manifest-templates/locust-worker-controller.yaml.tpl#L57:L63

benchmarks/benchmark/tools/latency-profile/manifest-templates/hf-secret.yaml.tpl
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: hf-token
+  namespace: ${namespace}
+data:
+  HF_TOKEN: ${hugging_face_token_b64}
```
annapendleton, line 7: This looks like automation for creating the k8s hardcoded secret pathway (an option that enables customers to use their own k8s secret instead of using secret manager) that uses k8s_hf_secret: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/1f4c96861ad166fd8c2c7c1847e45c8b187e362c/benchmarks/benchmark/tools/locust-load-inference/variables.tf#L201

Long term I think we should consider using the same implementation pathway as locust does for this (we can keep the creation automation as well), and also make sure we support both pathways (secret manager and k8s secret)

benchmarks/benchmark/tools/latency-profile/manifest-templates/namespace.yaml.tpl
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
```
annapendleton, line 2: This duplicates the work that the ai-on-gke benchmark/infra does

ref: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/1f4c96861ad166fd8c2c7c1847e45c8b187e362c/benchmarks/infra/stage-2/main.tf#L22

this causes conflicts if the namespace already exists, which is likely if a user comes through the the infra/ pathway rather than the Bring Your Own Cluster pathway. I think we'll need a cleaner story to support both pathways

benchmarks/benchmark/tools/latency-profile/manifest-templates/service-account.yaml.tpl
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
```
annapendleton, line 2: The infra/ creates a service account for the benchmarking tooling: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/1f4c96861ad166fd8c2c7c1847e45c8b187e362c/benchmarks/infra/stage-2/main.tf#L31

Long-term, I think we'll want to clean up the service account story for both pathways:
1. people who already created service account via infra/ automation
2. people who need to create a new service account since they didn't use infra/ automation (aka BYOCluster)

Bslabe123: This change was diagnostic, has since been removed

benchmarks/benchmark/tools/latency-profile/manifest-templates/namespace.yaml.tpl
Bslabe123: See the readme references this as well, removed

benchmarks/benchmark/tools/profile-generator/sample.tfvars
```diff
@@ -0,0 +1,108 @@
...
+credentials_config = {
+  kubeconfig = {
+    path = "~/.kube/config"
+  }
+}
+
+project_id = "tpu-vm-gke-testing"
+
+# Latency profile generator service configuration
+artifact_registry                          = "us-central1-docker.pkg.dev/tpu-vm-gke-testing/ai-benchmark"
```
achandrasekar, line 26: Let's not include real project id here and above. @Bslabe123

benchmarks/benchmark/tools/profile-generator/main.tf
```diff
@@ -0,0 +1,104 @@
...
+
+  timeouts {
+    create = "30m"
+    update = "40m"
+  }
+
+  disable_on_destroy = false
+}
+
+# CREATE NODEPOOLS
```
achandrasekar, line 53: Are we not creating node pools in the infra stage?

benchmarks/benchmark/tools/latency-profile/main.tf
```diff
@@ -0,0 +1,108 @@
...
+  manifest   = yamldecode(each.value)
+  timeouts {
+    create = "30m"
+  }
+}
+
+resource "null_resource" "cleanup_model_server" {
+  depends_on = [resource.kubernetes_manifest.deploy_latency_profile_generator]
+  provisioner "local-exec" {
+    command = "kubectl wait --for=condition=complete job/latency-profile-generator --timeout=-9600s && echo hello"
```
achandrasekar, line 103: Do we want to echo something more meaningful other than "hello"?

benchmarks/benchmark/tools/profile-generator/main.tf
Bslabe123: Not for all of the various accelerators we plan to use

benchmarks/benchmark/tools/latency-profile/main.tf
Bslabe123: You're too early! This is a placeholder

benchmarks/benchmark/tools/latency-profile/manifest-templates/latency-profile-generator.yaml.tpl
annapendleton: looks like this was addressed, resolving :)

benchmarks/benchmark/tools/locust-load-inference/main.tf
```diff
@@ -34,7 +34,7 @@ locals {
     [for data in split("---", templatefile(manifest_file, {
       artifact_registry              = var.artifact_registry
       namespace                      = var.namespace
-      inference_server_service       = var.inference_server_service
+      inference_server_service       = var.inference_server.service
```
annapendleton, line 37: @Bslabe123 this change will break locust automation - let's revert this file, since the latency profiler is a separate tool

benchmarks/benchmark/tools/profile-generator/main.tf
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  disable_on_destroy = false
+}
+
+#  ----- Manual Benchmarking -----
+
+module "profile-generator" {
+  count = var.targets.manual != null ? 1 : 0
+  source = "../latency-profile"
```
annapendleton, line 55: @Bslabe123 @achandrasekar wondering if we should organize this so that it's profile-generator top level, and the latency-profile engine folder within it, eg:

profile-generator/
     latency-profile/
     
This makes it clear via directory level that this is all one tool (eg. akin to how locust and dlio have their own top level file to represent the tooling)

Bslabe123: Added for this plus deduplicating the image building logic

annapendleton: that lgtm, ty ty!

annapendleton (approved):

Bslabe123 (approved):

PR #781, opened 2024-08-21
fix: use ray addon when creating GKE clusters

genlu2011:
This PR uses ray addon to install kuberay operator and removes the overhead to manage logging and monitoring. 
1. Upgrade the gke submodule version to **32.0.1** to use ray-addon to create GKE cluster.
2. Upgrade the k8s version to **1.30.3** to use ray addon.  
3. Removed the configuration related to fluent-bit image in Ray cluster module, since it is managed by ray addon now.
4. Removed the configuration to setup monitoring in kuberay-monitoring module.  
5. Removed installing kuberay-operator, kuberay-logging module in rag/ray application.


applications/ray/main.tf
```diff
@@ -147,24 +151,6 @@ module "kuberay-workload-identity" {
...
-
-module "kuberay-logging" {
-  source    = "../../modules/kuberay-logging"
-  providers = { kubernetes = kubernetes.ray }
-  namespace = local.kubernetes_namespace
-
-  depends_on = [module.namespace]
-}
-
 module "kuberay-monitoring" {
```
andrewsykim, line 154: I think we can remove kuberay-monitoring too if we enable the monitorign feature in the addon

applications/ray/main.tf
```diff
@@ -147,24 +151,6 @@ module "kuberay-workload-identity" {
   depends_on                      = [module.namespace]
 }
 
-module "kuberay-operator" {
-  source            = "../../modules/kuberay-operator"
```
andrewsykim, line 151: Can we delete the source modules too?

applications/ray/main.tf
```diff
@@ -147,24 +151,6 @@ module "kuberay-workload-identity" {
...
-  providers         = { helm = helm.ray, kubernetes = kubernetes.ray }
-  name              = "kuberay-operator"
-  create_namespace  = true
-  namespace         = local.kubernetes_namespace
-  project_id        = var.project_id
-  autopilot_cluster = local.enable_autopilot
-}
-
-module "kuberay-logging" {
-  source    = "../../modules/kuberay-logging"
```
andrewsykim, line 161: also delete the module itself

applications/ray/main.tf
genlu2011: This module now also handles use grafana in ray dashboard, which is not in monitoring feature in the addon.

applications/ray/main.tf
genlu2011: deleted

applications/ray/main.tf
genlu2011: deleted

imreddy13 (approved):

PR #785, opened 2024-08-27
Fixing issue with package notebook in jupyterhub image.

german-grandas:
- Refactoring of code so it can build and install directly the authentication package instead of getting it from github.
 - Installing notebook as a separate dependency
 - Letting pip install the jupyterhub version according to the notebook package.


modules/jupyter/authentication/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM jupyterhub/k8s-hub:3.3.8
+
+ADD ./authenticator/ ./authenticator
+
+USER root
+RUN chmod -R u+w ./authenticator
+
+RUN pip3 install notebook
```
chiayi, line 8: What's the reason for installing the notebook for the hub image?

modules/jupyter/authentication/authenticator/setup.py
```diff
@@ -23,10 +23,10 @@
     long_description="long description here",
     long_description_content_type="text/markdown",
     url="https://github.com/chiayi/ai-on-gke/tree/main/ray-on-gke/user/jupyterhub/auth",
-    packages=['gcpiapjwtauthenticator'],
+    packages=setuptools.find_packages(),
     license='Apache 2.0',
     install_requires=[
-        "jupyterhub>=4.1.0",
+        "jupyterhub",
```
chiayi, line 29: I believe version 4.1.0 contains a fix for a known vulnerability and thats why we have `>=4.1.0"

chiayi (commented):
Hi [german-grandas](https://github.com/german-grandas), thanks for contributing to the repo! Just have a few questions and comments.

german-grandas:
Closing PR, the bug is fixed!

PR #789, opened 2024-08-28
fix: add vpc cleanup cloudbuild yaml

genlu2011:
Fix the script to delete leftover VPCs, which only deletes firewall before the fix. 
However, to delete a VPC, the linked subnets, firewall, address, peering, route have to be deleted first.
 
In addition, instead of using inline YAML in gcp console to define the cloud build job, move the script to this repo. 
1. Safer and protect us from accidentally delete the inline script. 
2. Requires code review to change the definition in production.


yiyinglovecoding:
Are you going to update the cloud build trigger to use the script instead?

yiyinglovecoding (approved):

PR #790, opened 2024-08-29
fix: increase wait time for ray head

genlu2011:
[skip ci] 
During investigate the flakiness of our CI, I found the root cause for majority of ray application's flakiness is the readinessProbe.
Today in rag application test:
1. We use kubectl wait --all pods -n ml-$SHORT_SHA-$_BUILD_ID-ray --for=condition=Ready --timeout=1200s to wait for the Ray head to be ready. When ray head is ready, it serves the traffic on address: [127.0.0.1:6379](http://127.0.0.1:6379/)
2. However, the readinessProbe is defined as:
    readinessProbe:
      exec:
        command:
        - bash
        - -c
        - wget -T 2 -q -O- http://localhost:52365/api/local_raylet_healthz | grep
          success && wget -T 2 -q -O- http://localhost:8265/api/gcs_healthz | grep
          success
3. The prober only probes the raylet and gcs, but not the ray service on port 6379

Therefore, in our CI tests, we see that when a job is submitted in ray application test, the connection is broken due to the server is not ready.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -102,8 +102,8 @@ steps:
 
         # Make sure pods are running
         kubectl wait --all pods -n ml-$SHORT_SHA-$_BUILD_ID-ray --for=condition=Ready --timeout=1200s
-        # Wait for pods to be stable
-        sleep 5s
+        # Ray head's readinessProbe is not probing the head service today. Therefore the wait for ready above is not reliable.
+        sleep 45s
         kubectl port-forward -n ml-$SHORT_SHA-$_BUILD_ID-ray service/ray-cluster-kuberay-head-svc 8265:8265 &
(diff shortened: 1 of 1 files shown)


genlu2011:
[skip ci]

spencer-p:
Could we change the readiness probe to require :6379 to be live instead of adding a 40s sleep?

genlu2011:
> Could we change the readiness probe to require :6379 to be live instead of adding a 40s sleep?

I have created a ticket to track that. The module used to create ray cluster is owned by another repo.

genlu2011:
Added in https://github.com/GoogleCloudPlatform/ai-on-gke/pull/781, which also touched cloudbuild.yaml.

PR #793, opened 2024-08-31
Adding support vLLM openai entrypoint benchmarking script

Edwinhr716:
Currently the benchmarking script only works with the regular entrypoint of vllm. Since the benchmarking suite builds uses openai entrypoint instead, changed it to support that entrypoint
.

### DIFF
--- a/benchmarks/benchmark/tools/profile-generator/container/latency_throughput_curve.sh
+++ b/benchmarks/benchmark/tools/profile-generator/container/latency_throughput_curve.sh
@@ -25,6 +25,6 @@ for request_rate in $(echo $REQUEST_RATES | tr ',' ' '); do
   output_file="latency-profile-${timestamp}.txt"
   python3 benchmark_serving.py   --host="$IP"   --port="$PORT"   --model="$TOKENIZER" --dataset=ShareGPT_V3_unfiltered_cleaned_split.json   --tokenizer="$TOKENIZER" --request-rate=$request_rate --backend="$BACKEND" --num-prompts=$((request_rate * 30)) --max-input-length=$INPUT_LENGTH --max-output-length=$OUTPUT_LENGTH > $output_file
   cat $output_file
-  sleep 5 # wait 5 minutes before next run
+  sleep 5 # wait 5 seconds before next run
 done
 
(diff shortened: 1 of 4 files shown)


benchmarks/benchmark/tools/profile-generator/container/latency_throughput_curve.sh
```diff
@@ -23,5 +23,8 @@ for request_rate in $(echo $REQUEST_RATES | tr ',' ' '); do
   # TODO: Check if profile already exists, if so then skip
   timestamp=$(date +"%Y-%m-%d_%H-%M-%S")
   output_file="latency-profile-${timestamp}.txt"
-  python3 benchmark_serving.py   --host="$IP"   --port="$PORT"   --dataset=ShareGPT_V3_unfiltered_cleaned_split.json   --tokenizer="$TOKENIZER" --request-rate=$request_rate --backend="$BACKEND" --num-prompts=$((request_rate * 30)) --max-input-length=$INPUT_LENGTH --max-output-length=$OUTPUT_LENGTH > $output_file
-done
\ No newline at end of file
+  python3 benchmark_serving.py   --host="$IP"   --port="$PORT"   --model="$TOKENIZER" --dataset=ShareGPT_V3_unfiltered_cleaned_split.json   --tokenizer="$TOKENIZER" --request-rate=$request_rate --backend="$BACKEND" --num-prompts=$((request_rate * 30)) --max-input-length=$INPUT_LENGTH --max-output-length=$OUTPUT_LENGTH > $output_file
+  cat $output_file
+  sleep 5 # wait 5 minutes before next run
```
annapendleton, line 28: IIRC, sleep is based in seconds, so this would be 5 seconds?

annapendleton (approved):

benchmarks/benchmark/tools/profile-generator/container/latency_throughput_curve.sh
Edwinhr716: correct, fixed the comment

vivianrwu (commented):
For jetstream, can you change `max_tokens: 1` to `max_tokens: output_len` under the `benchmark_serving.py` script? Other than that, lgtm for jetstream side changes!

PR #794, opened 2024-09-03
[FIX] Single GPU training job update

xiangshen-dk:
Due to [Keras is upgraded to 3 after TF 2.16](https://blog.tensorflow.org/2024/03/whats-new-in-tensorflow-216.html), the single GPU job is broken. 

This PR is to fix the issue.


alizaidis (approved):
LGTM

PR #801, opened 2024-09-07
Allow benchmark to write json output

richardsliu:
Modify the serving benchmark to write json output similar to Jetstream benchmark: https://github.com/richardsliu/JetStream/blob/vllm-benchmark/benchmarks/benchmark_serving.py#L648

This will allow the benchmark to integrate with XLML dashboard.


achandrasekar (approved):
Looks good. Thanks for adding this! Makes it easier to parse the results.

PR #803, opened 2024-09-09
feat: add deployment of nvidia nim

laurentgrangeau:
Adding deployment of NVIDIA NIMs on the ML platform

### DIFF
--- a/best-practices/ml-platform/terraform/features/nim/outputs.tf
+++ b/best-practices/ml-platform/terraform/features/nim/outputs.tf
@@ -1,0 +1,3 @@
+output "inference-url" {
+  value = ""
+}
--- a/best-practices/ml-platform/terraform/features/nim/outputs.tf
+++ b/best-practices/ml-platform/terraform/features/nim/outputs.tf
@@ -1,3 +1,3 @@
-output "inference-url" {
+output "inference_url" {
   value = ""
 }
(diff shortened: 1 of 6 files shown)


best-practices/ml-platform/terraform/features/nim/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "inference-url" {
+  value = ""
```
arueth, line 2: If this isn't going to be populated with a value, can you remove it?

arueth:
For the repo, we try and follow the [Terraform on Google Cloud: Best practices for general style and structure](https://cloud.google.com/docs/terraform/best-practices/general-style-structure), could you take a look at it and update the PR. Mainly around the [Adopt a naming convention](https://cloud.google.com/docs/terraform/best-practices/general-style-structure#naming-convention) and using underscore to delimit multiple words.

Although not required, it is good practice to break out standard attributes and block attributes as well as alphabetize attributes in resources and data sources to make merges and updates from multiple people cleaner. 

Also please ensure that you run `terraform fmt` on the code as this will be enforce when merging to main.

arueth (commented):
Also see my prior comment.

roberthbailey:
How different is this from the nim guide that we merged into https://github.com/GoogleCloudPlatform/ai-on-gke/tree/main/tutorials-and-examples/nvidia-nim?

arueth:
Looks to be similar, except done with Terraform. We are consuming this in the `ml-platform`, so it will be referenced internally primarily.

PR #804, opened 2024-09-11
Add server metrics promql scraping

annapendleton:
Add initial implementation for scraping server side metrics for analysis.

Initial implementation includes support for vLLM and Jetstream.

Validated end to end, example output at end of output file:
```
Metric Name: vllm:gpu_cache_usage_perc
Mean: 0.6996527777777777
Median: 0.8744212962962963
Min: 0.05439814814814814
Max: 0.9953703703703703
P90: 0.9936342592592593
P99: 0.9951967592592592
Metric Name: vllm:num_requests_waiting
Mean: 25.75
Median: 2.5
Min: 0
Max: 98
P90: 70.10000000000002
P99: 95.20999999999998
```

### DIFF
--- a/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
+++ b/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
@@ -609,8 +609,7 @@ if __name__ == "__main__":
   )
   parser.add_argument(
       "--scrape-server-metrics",
-      type=bool,
-      default=False,
+      action="store_true",
       help="Whether to scrape server metrics.",
   )
   cmd_args = parser.parse_args()
--- a/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
+++ b/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
@@ -314,7 +314,7 @@ def metrics_to_scrape(backend: str) -> List[str]:
   else:
     return []
 
-def print_metrics(metrics: List[str], duration: str, backend: str) -> None:
+def print_metrics(metrics: List[str], duration: float, backend: str):
   # Creates a credentials object from the default service account file
   # Assumes that script has appropriate default credentials set up, ref:
   # https://googleapis.dev/python/google-auth/latest/user-guide.html#application-default-credentials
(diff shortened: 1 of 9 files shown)


benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -302,6 +306,54 @@ def save_json_results(args: argparse.Namespace, benchmark_result):
...
+  else:
+    return []
+
+def print_metrics(metrics: List[str], duration: str, backend: str) -> None:
+  # Creates a credentials object from the default service account file
+  # Assumes that script has appropriate default credentials set up, ref:
+  # https://googleapis.dev/python/google-auth/latest/user-guide.html#application-default-credentials
+  credentials, project_id = google.auth.default()
+  # Prepare an authentication request - helps format the request auth token
+  auth_req = google.auth.transport.requests.Request()
```
achandrasekar, line 323: Does authentication work by default when using the automation? If not, let's put metrics scraping behind a flag so that it does not error for people who want to run the rest of the script.

benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -302,6 +306,54 @@ def save_json_results(args: argparse.Namespace, benchmark_result):
...
+      # Configure respective query
+      url='https://monitoring.googleapis.com/v1/projects/%s/location/global/prometheus/api/v1/query' % (project_id)
+      headers_api = {'Authorization': 'Bearer ' + credentials.token}
+      params = {'query': query}
+      response = requests.get(url=url, headers=headers_api, params=params)
+
+      # handle response
+      if response.ok:
+        if response["status"] == "success":
+          print("%s: %s" % (query_name, response["data"]["result"][0]["value"][1]))
```
achandrasekar, line 351: Now we have the ability to save results to a json file. It would be good to add this to the json file as well since it is easier to parse programatically.

benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
annapendleton: Not yet, I'll implement the flag! And also update the README with the necessary instructions.

benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
annapendleton: That makes sense, I'll add that in!

benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
annapendleton: Done!

benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
annapendleton: Done!

benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -545,5 +607,11 @@ def main(args: argparse.Namespace):
           " the form of a string."
       ),
   )
+  parser.add_argument(
+      "--scrape-server-metrics",
+      type=bool,
```
achandrasekar, line 612: For other arguments that are booleans like this we are using `action=store_true` like `--save-json-result` or `--use-dummy-text` where specifying just the flag is enough. But it seems like setting it to a bool allows us to easily specify values at runtime. Maybe we can switch the others to bool as well if we want to follow the same pattern. Not a blocker. But something to address.

achandrasekar (approved):

benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
annapendleton: I double checked to see if using the type bool worked with false - turns out python ignored what was set if the flag was passed in. I was able to fix this doing something similar to what we have in the locust run.sh file.

Ref: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/0294ba355e8d0809fc4c37e7dcd0365358521af4/benchmarks/benchmark/tools/locust-load-inference/locust-docker/locust-tasks/run.sh#L19

PR #805, opened 2024-09-11
Adding a v6e network optimization daemonset to ai-on-gke repository.

NoahJohnson1:
This checks in a new DaemonSet composed primarily of an initContainer running a network optimization bash script.

The script itself does a number of things that Google has measured to improve network performance. This includes "IRQ spreading", which ensures that each vnic is having interrupts processed on a distinct set of cores, various TCP settings, and enabling larger gve driver packet buffers, which improve performance when a large MTU is used.


thebinaryone1 (approved):

thebinaryone1 (commented):
/gcbrun

PR #806, opened 2024-09-11
fix: upgrade gke module to avoid TPGv5.44.0 with Autopilot

genlu2011:
Upgrade gke submoduel version to patch the fix in latest release
https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/releases/tag/v33.0.0

### DIFF
--- a/infrastructure/versions.tf
+++ b/infrastructure/versions.tf
@@ -18,7 +18,8 @@ terraform {
       source = "hashicorp/google"
     }
     google-beta = {
-      source = "hashicorp/google-beta"
+      source  = "hashicorp/google-beta"
+      version = ">= 5.40.0, != 5.44.0, !=6.2.0"
     }
     helm = {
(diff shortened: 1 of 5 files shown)


infrastructure/versions.tf
```diff
@@ -18,7 +18,8 @@ terraform {
       source = "hashicorp/google"
     }
     google-beta = {
-      source = "hashicorp/google-beta"
+      source  = "hashicorp/google-beta"
+      version = ">= 5.40.0, != 5.44.0, !=6.2.0"
```
andrewsykim, line 22: Leave a comment for while these versions are excluded

andrewsykim (approved):
Minor nit about leaving a comment, otherwise LGTM

infrastructure/versions.tf
genlu2011: Added

PR #816, opened 2024-09-16
Update references to Hugging Face DLC for TGI

alvarobartt:
## Description

This PR updates a bunch of outdated references to Text Generation Inference (TGI) container, as previously those were pointing to the GitHub Container Registry where Hugging Face publishes those containers i.e. ghcr.io/huggingface/text-generation-inference; but since the Hugging Face DLCs on Google Cloud were recently released publicly (see [`Google-Cloud-Containers`](https://github.com/huggingface/Google-Cloud-Containers)), now those references have been updated to point to the DLCs hosted on Google Cloud's Artifact Registry.

cc @brandonroyal

### DIFF
--- a/benchmarks/inference-server/text-generation-inference/manifest-templates/text-generation-inference.tftpl
+++ b/benchmarks/inference-server/text-generation-inference/manifest-templates/text-generation-inference.tftpl
@@ -51,7 +51,7 @@ spec:
         - name: text-generation-inference
           ports:
             - containerPort: 80
-          image: "us-docker.pkg.dev/deeplearning-platform-release/gcr.io/huggingface-text-generation-inference-cu121.2-2.ubuntu2204.py310"
+          image: "ghcr.io/huggingface/text-generation-inference:1.4.2"  # TODO(annapendleton,alvarobartt): update to Hugging Face DLC once verified
           args: ["--model-id", "${model_id}", "--num-shard", "${gpu_count}", "--max-concurrent-requests", "${max_concurrent_requests}"]
           env:
 %{ for hugging_face_token_secret in hugging_face_token_secret_list ~}
--- a/benchmarks/inference-server/text-generation-inference/manifest-templates/text-generation-inference.tftpl
+++ b/benchmarks/inference-server/text-generation-inference/manifest-templates/text-generation-inference.tftpl
@@ -51,7 +51,7 @@ spec:
         - name: text-generation-inference
           ports:
             - containerPort: 80
-          image: "ghcr.io/huggingface/text-generation-inference:1.4.2"
+          image: "us-docker.pkg.dev/deeplearning-platform-release/gcr.io/huggingface-text-generation-inference-cu121.2-2.ubuntu2204.py310"
           args: ["--model-id", "${model_id}", "--num-shard", "${gpu_count}", "--max-concurrent-requests", "${max_concurrent_requests}"]
           env:
 %{ for hugging_face_token_secret in hugging_face_token_secret_list ~}
--- a/modules/inference-service/main.tf
+++ b/modules/inference-service/main.tf
@@ -89,7 +89,7 @@ resource "kubernetes_deployment" "inference_deployment" {
           }
         }
         container {
-          image = "ghcr.io/huggingface/text-generation-inference:1.1.0"
+          image = "us-docker.pkg.dev/deeplearning-platform-release/gcr.io/huggingface-text-generation-inference-cu121.2-2.ubuntu2204.py310"
           name  = "mistral-7b-instruct"
 
           port {
--- a/tutorials-and-examples/genAI-LLM/deploying-mistral-7b-instruct-L4gpus/README.md
+++ b/tutorials-and-examples/genAI-LLM/deploying-mistral-7b-instruct-L4gpus/README.md
@@ -8,7 +8,7 @@ Learn how to serve the Mistral 7B instruct v0.1 chat model on GKE using just 1 x
 *   GPU Quota: Confirm you have the quota for at least one L4 GPU in your Google Cloud account.
 *   Model Access: Secure access to the Mistral 7B model by agreeing to the terms on Hugging Face, which typically involves creating an account and accepting the model's use conditions.
 *   Ensure you currently have installed a stable version of Transformers, 4.34.0 or newer.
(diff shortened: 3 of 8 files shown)


benchmarks/inference-server/text-generation-inference/manifest-templates/text-generation-inference.tftpl
```diff
@@ -51,7 +51,7 @@ spec:
         - name: text-generation-inference
           ports:
             - containerPort: 80
-          image: "ghcr.io/huggingface/text-generation-inference:1.4.2"
+          image: "us-docker.pkg.dev/deeplearning-platform-release/gcr.io/huggingface-text-generation-inference-cu121.2-2.ubuntu2204.py310"
```
annapendleton, line 54: @alvarobartt this is a major upgrade for TGI (eg. 1.4.2 -> 2.2 per docs here, https://huggingface.co/blog/llama31-on-vertex-ai#3-register-the-meta-llama-31-405b-model-on-vertex-ai)

OOC, how have you tested this for the benchmarking suite?

annapendleton (changes requested):

benchmarks/inference-server/text-generation-inference/manifest-templates/text-generation-inference.tftpl
alvarobartt: Hi @annapendleton, yes indeed the upgrade comes with performance improvements and support for more architectures; and indeed TGI < 2.0 is no longer the recommended approach; anyway those models should work normally with either version.

Also w.r.t. benchmarking we've done extensive benchmarking on TGI globally yes! Or do you mean with any suite you may have?

Happy to help with more information on any of those if needed! 🤗

alvarobartt: Alternatively, TGI 1.4 is also available within the Hugging Face DLCs on Google Cloud at us-docker.pkg.dev/deeplearning-platform-release/gcr.io/huggingface-text-generation-inference-cu121.1-4.ubuntu2204.py310; but as mentioned, not recommended, anyway worst case scenario we can pin that URI for the mentioned example if needed.

annapendleton: Ah I just mean have you tested that the ai-on-gke/tree/main/benchmarks/inference-server/text-generation-inference works with the upgraded image :) I wasn't sure if there were any flag / use changes between the major versions that would break what is already supported in ai-on-gke/benchmarks suite. We don't have any automated tests so it requires manual validation. From what I've read, it sounds like this should work as a drop in replacement?

alvarobartt: Oh yes sorry I didn't get that! Yes, should work as is, no need to change anything else! Anyway I believe we can still re-run the benchmarking suite on your end, is that something you can do, should we run that on our end? cc'ing also @brandonroyal here just in case there's anything we can help!

annapendleton: @alvarobartt I won't have time to validate this as I'll be leaving for vacation in a few days - is there a desired deadline on when we upgrade the ai-on-gke/benchmarks to this image?

If it's not urgent, I would recommend we remove the upgrade in ai-on-gke/benchmarks from this PR so you can submit for the other related tutorials, and our team can take this task on separately to upgrade and validate (and also thinking we might want to parameterize this so users can specify which image they want to benchmark, and make this image the default image).

alvarobartt: Fair, I'll leave this comment open so that we can come back to this once you're back from vacation maybe and see if we can also update the container URI used for benchmarking? Thanks for the information 🤗

annapendleton: Sounds good to me! Thanks for your flexibility. I also opened an issue for tracking, so definitely feel free to close this comment and submit this PR for the other related tutorials :)

ref: https://github.com/GoogleCloudPlatform/ai-on-gke/issues/821

annapendleton (approved):

annapendleton:
@alvarobartt I also added you as a collaborator so you can test + merge this yourself :)

alvarobartt:
Hi here @annapendleton, so the CI is failing, but not sure that's something I can fix or related to any of the changes within this PR, could you help me debug the issue further if related to the PR? 🤗

annapendleton:
> Hi here @annapendleton, so the CI is failing, but not sure that's something I can fix or related to any of the changes within this PR, could you help me debug the issue further if related to the PR? 🤗

From what I can see in the logs, it looks like the test cluster isn't present. We don't use these tests ourselves for the benchmarking, I think we should add someone else who has a better sense of if it's safe to bypass this test or not.

annapendleton:
@imreddy13 could you help us determine if this is a true test failure?

alvarobartt:
Hi here @annapendleton @brandonroyal is there any update with regards to this PR? Thanks in advance 🤗

annapendleton:
@alvarobartt tests are passing, let me know if you still want to merge :)

alvarobartt:
Hi here @annapendleton thanks for handling!

So now the latest released DLC for TGI is 2.3 (not 2.2 as in this PR) do you think makes sense to just update that now and re-run the tests?

Also what's the strategy you usually follow to keep the documentation examples point to the latest containers in these cases? Or should we just open a PR here updating those on each release and then waiting for the tests to pass?

Thanks in advance 🤗

annapendleton:
@alvarobartt you'll have to open a new PR to keep the containers updated with each release you'll want to upgrade to for the DLC used in these example workloads. It'll be the same process as done here.

If you want to upgrade, we'll have to rerun the tests anyways, so let's do that in a new PR :) I merged this PR for the current container image.

raushan2016:
This looks very similar to the change below which has issue where the new image doesn't work for LLAMA multiple GPU.
https://github.com/GoogleCloudPlatform/kubernetes-engine-samples/issues/1581

@annapendleton

alvarobartt:
Hi here @annapendleton indeed see this PR with the fix for the current repository (on a previously opened PR updating the TGI DLC)

https://github.com/GoogleCloudPlatform/ai-on-gke/pull/852

PR #817, opened 2024-09-16
fix: remove gke cluster version

genlu2011:
The pinged version has reached to REGULAR channel. 
The default GKE version used in our TF is 1.30, which will pick the highest version in REGULAR channel.

### DIFF
--- a/applications/rag/main.tf
+++ b/applications/rag/main.tf
@@ -79,10 +79,7 @@ module "infra" {
   enable_gpu        = true
   gpu_pools         = var.gpu_pools
   ray_addon_enabled = true
-  # TODO(genlu): remove channel and k8s_version after ray addon is in REGULAR channel
-  release_channel    = "RAPID"
-  kubernetes_version = "1.30.3-gke.1969000"
-  depends_on         = [module.project-services]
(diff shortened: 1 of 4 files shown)


applications/rag/main.tf
```diff
@@ -79,9 +79,6 @@ module "infra" {
   enable_gpu        = true
   gpu_pools         = var.gpu_pools
   ray_addon_enabled = true
-  # TODO(genlu): remove channel and k8s_version after ray addon is in REGULAR channel
-  release_channel    = "RAPID"
-  kubernetes_version = "1.30.3-gke.1969000"
```
andrewsykim, line 84: Does the default regular release channel version have the bug fix required to use addon with resource quotas?

spencer-p: Regular deafult (1.30.3-gke.1639000) doesn't  set resource requests for the fluentbit container. 1.30.3-gke.1969001 does. Its equivalent in rapid, 1.30.3-gke.1969000, is the rapid channel upgrade target.

Rapid default is 1.31.0-gke.1506000 which also has the resource requests patch.

andrewsykim (approved):

PR #830, opened 2024-09-20
Remove unneeded depdendencies in LPG container image

Bslabe123:


benchmarks/benchmark/tools/profile-generator/container/Dockerfile
```diff
@@ -1,9 +1,8 @@
-FROM nvidia/cuda:12.1.0-devel-ubuntu22.04 AS dev
+FROM python:3.9.20-slim-bookworm as dev
 
 RUN apt-get update -y \
     && apt-get install -y python3-pip git vim curl wget
 RUN pip3 install --upgrade pip
-RUN pip install packaging torch transformers
```
achandrasekar, line 6: How would we tokenize without the transformers library?

Bslabe123: From what I understand [we're getting our tokenizer](https://github.com/GoogleCloudPlatform/ai-on-gke/blob/main/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py#L39) from `transformers`

Bslabe123: [We still install transformers here](https://github.com/GoogleCloudPlatform/ai-on-gke/pull/830/files#diff-227c9ee0815e0858c88d9533a0f315c4db1384eb28da9dda60f883ad2e2bf35eR28)

achandrasekar: Ok, that makes sense. It would be good to update that transformers library to a later version so it has support for newer model tokenizers. I think 4.42+ is available now.

achandrasekar (approved):

PR #833, opened 2024-09-25
Make the benchmark time configurable and set default to 2 minutes

achandrasekar:
This change makes the benchmark time configurable so it makes it easy to tweak it and identify when we reach steady state. This change also increases the default time from 30 seconds to 2 minutes based on experimental data.


Bslabe123 (approved):
LGTM

PR #836, opened 2024-09-25
Rely on script to gather data, not Prometheus

Bslabe123:
The metrics the benchmarking script gathers will be the closest to what an end user would see, as such we should use these for generating profiles. `start_time` has been temporarily removed since more investigation is needed to correctly implement this, this will be re-added in a follow up PR/

### DIFF
--- a/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
+++ b/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
@@ -300,7 +300,7 @@ def save_json_results(args: argparse.Namespace, benchmark_result, server_metrics
     "config": {
       "model": args.model,
       "model_server": args.backend,
-      "start_time": start_dt_proto.ToJsonString(),
+      # "start_time": start_dt_proto.ToJsonString(), TODO: Marshall this such that the unmarshaller doesnt error
     },
     "summary_stats": {
       "stats": [{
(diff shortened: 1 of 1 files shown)


benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -300,12 +300,52 @@ def save_json_results(args: argparse.Namespace, benchmark_result, server_metrics
     "config": {
       "model": args.model,
       "model_server": args.backend,
-      "start_time": start_dt_proto.ToJsonString(),
+      # "start_time": start_dt_proto.ToJsonString(), TODO: Marshall this such that the unmarshaller doesnt error
```
achandrasekar, line 303: It would be good to have some time here even if it is not entirely accurate so we have some way to know when it was run. Maybe datetime.now.strftime()?

benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -404,17 +438,41 @@ def print_metrics(metrics: Dict[str, Optional[str]], duration: float, backend: s
...
+def get_stats_for_set(name, description, points):
+  avg = np.mean(points)
+  median = np.median(points)
+  sd = np.std(points)
+  min = np.min(points)
+  max = np.max(points)
+  p90 = np.percentile(points, 90)
+  p99 = np.percentile(points, 99)
+
+  print(f"Average {description}:" f" {avg:.2f}")
```
achandrasekar, line 459: Since we are adding all the p90, mean, sd data in the summary stat, can we keep this section light and report only mean like before? Other option is to club them into separate sections like we do in summary stats so it is more readable. wdyt?

achandrasekar: Sounds good. Even in json it would be good to not repeat it in two places. So, thanks for addressing this.

achandrasekar (approved):

PR #839, opened 2024-09-29
Fix throughput to be in output tokens per second

achandrasekar:
Throughput was changed to be in rps from output tokens per second. We don't want this to be the default. So, fixing that.

### DIFF
--- a/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
+++ b/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
@@ -518,11 +518,14 @@ def main(args: argparse.Namespace):
   print(f"Total time: {benchmark_time:.2f} s")
   print(f"Requests/min: {60 * args.num_prompts / benchmark_time:.2f}")
   benchmark_result['benchmark_time'] = benchmark_time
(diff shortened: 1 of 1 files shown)


benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -518,11 +518,14 @@ def main(args: argparse.Namespace):
...
   print(f"Requests/min: {60 * args.num_prompts / benchmark_time:.2f}")
   benchmark_result['benchmark_time'] = benchmark_time
-  benchmark_result['throughput'] = (args.num_prompts / benchmark_time)
+  benchmark_result['throughput_rps'] = (args.num_prompts / benchmark_time)
 
   total_output_tokens = np.sum([output_len for _, output_len, _ in
                                 REQUEST_LATENCY])
-  output_tokens_per_min = 60 * total_output_tokens / benchmark_time
+  output_tokens_per_second = total_output_tokens / benchmark_time
+  benchmark_result['throughput'] = output_tokens_per_second
```
Bslabe123, line 526: Not blocking but maybe worth it to rename `throughput` -> `throughput_tps` since throughput is typically denoted in requests per second?

Bslabe123 (approved):

PR #844, opened 2024-10-04
Made prompt dataset configurable

Bslabe123:


benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -32,6 +33,14 @@
 CLIENT_TIMEOUT_SEC = 3 * 60 * 60
 NEW_TEXT_KEY = "\nOutput:\n"
 
+def save_dataset(dataset_url) -> str:
+  file_name = os.path.basename(dataset_url)
+  response = requests.get(dataset_url)
+  print(f"downloading {dataset_url}")
```
achandrasekar, line 39: There are a couple of issues I see with downloading this in at run time?

1. Reliability of fetching them at run time is an issue - there are times where the url might not be reachable. In some cases the dataset can be multiple GBs like the OpenOrca dataset.
2. Datasets don't have the data in the same format. So, if we are passing in URL for another dataset, that might need separate logic to parse it which we might not have. So, we need a way to address that as well.

achandrasekar (approved):

PR #848, opened 2024-10-09
Update openssh version in jupyter image

vicentefb:
The most important fix is to update the OpenSSH package in the Dockerfile. You can usually do this by using the package manager within your Dockerfile (e.g., apt-get update && apt-get install -y openssh-server for Debian/Ubuntu).

### DIFF
--- a/modules/jupyter/jupyter_image/notebook_image/Dockerfile
+++ b/modules/jupyter/jupyter_image/notebook_image/Dockerfile
@@ -1,3 +1,11 @@
 FROM jupyter/tensorflow-notebook:python-3.10
+USER root
 COPY requirements.txt ./requirements.txt
+RUN apt-get update && apt-get install -y --no-install-recommends openssh-server
+# Create a virtual environment
+RUN python3 -m venv /opt/venv
+
+# Activate the virtual environment
+ENV PATH="/opt/venv/bin:$PATH"
+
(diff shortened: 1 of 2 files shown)


modules/jupyter/jupyter_image/notebook_image/Dockerfile
```diff
@@ -1,3 +1,11 @@
 FROM jupyter/tensorflow-notebook:python-3.10
+USER root
 COPY requirements.txt ./requirements.txt
+RUN apt-get update && apt-get install -y --no-install-recommends openssh-server
+# Create a virtual environment
+RUN python3 -m venv /opt/venv
```
chiayi, line 6: What benefit does creating a virtual env do here?

vicentefb: This approach addresses the following pip warning which was triggered due to `USER root`: 

`Step` #1: WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv`

chiayi:
LGTM!

chiayi:
Discussed offline, v1.1 does not have the the vulnerability that requires updating the openssh version.

Issue #858, opened 2024-10-25
[Ray TPU Webhook] Pod Informer inconsistency for large RayCluster sizes


ryanaoleary:
The Ray TPU webhook sometimes fails to assign unique `TPU_WORKER_ID`s to all workers in a slice, resulting in the Jax initialization timing out. In manual testing this only occurs with large slice sizes (i.e. a v6e-256) around ~20% of the time (2 out of 10 trials) and is not reproduced in unit tests. I've previously tested the webhook with up to 4x4x8 v4 slices and did not observe this issue. The error is consistently fixed by re-creating the RayCluster. Initial debugging yielded the following results:
1. I initially thought the bug was due to a race condition with multiple concurrent mutation requests, but adding a `sync.Mutex` lock for the Pod mutation yielded no change. Additionally, mutating admission webhooks appear to execute requests [sequentially](https://open-policy-agent.github.io/gatekeeper/website/docs/mutation-background/#:~:text=A%20key%20difference%20between%20mutating,have%20no%20other%20side%20effects.), although I'm unclear whether this behavior is specific to Gatekeeper.
2. Log statements show that the error occurs when an admission request calculates the next `TPU_WORKER_ID` with a stale Informer cache (the PodLister does not show the updated `TPU_WORKER_ID` from the previous request) and re-assigns a worker ID.
3. Adding an additional `WaitForCacheSync` has no effect and checking `cache.WaitForCacheSync(stopCh, podInformer.HasSynced)` returns true even when the error occurs.
4. Adding a sufficiently long `time.Sleep()` after the Mutate function solves the issue but adds significant latency.

It appears the bug is a result of a slow update time to the PodInformer cache for a large number of KubeRay worker Pods. I'm currently investigating a better method to ensure the PodInformer retains an up-to-date list of Pods without resorting to calling `time.Sleep()`.

ryanaoleary:
cc: @andrewsykim

PR #859, opened 2024-10-25
Fix Webhook Assigning Identical `TPU_WORKER_ID`s

ryanaoleary:
This PR adds a `sync.WaitGroup` object and integer `waiting` var to each `TPUWebhookServer` object. All goroutine calls to `mutatePod` now `Wait()` before listing from the PodInformer cache, or timeout after 1 second, and then increment the waiting var and call `wg.Add(1)`. `wg.Done()` is called by the `AddFunc` `EventHandler`, which indicates the PodInformer has updated with the last Pod admitted by that webhook replica. This ensures the PodInformer cache is available and updating prior to listing from the cache. To support multiple webhook replicas, `wg.Done()` is only called if the int `waiting` var on the `TPUWebhookServer` object is greater than 1. This allows the webhook to block on previous `TPUWebhookServer.Mutate` calls until the PodInformer cache updates at least once.  I also added error checking for identical `TPU_WORKER_ID`s being assigned within the same slice (as opposed to just letting the Jax initialization time out).

Testing:
- [x] Unit Tests
- [x] Manual Tests: tested with a v6e-256 Ray TPU worker group for  1 and 3 webhook replicas respectively 

Related Issue #: [858](https://github.com/GoogleCloudPlatform/ai-on-gke/issues/858)

### DIFF
--- a/ray-on-gke/tpu/kuberay-tpu-webhook/main.go
+++ b/ray-on-gke/tpu/kuberay-tpu-webhook/main.go
@@ -754,14 +754,27 @@ func init() {
 // addPod allows next goroutine to start once the webhook PodInformer cache updates
 func (t *TPUWebhookServer) addPod(obj interface{}) {
 	// It's not guaranteed the webhook replica that admitted the Pod for this event is the same as the current caller (i.e. wg could be 0).
-	// It's sufficient to wait until the cache has updated once to unblock the next Mutate call.
-	if t.waiting > 0 {
-		defer t.wg.Done()
-	}
-
-	// Log the Pod being admitted
+	// It's sufficient to wait until the cache has updated once with a TPU pod to unblock the next Mutate call.
 	pod := obj.(*corev1.Pod)
 	klog.V(1).InfoS("addPod", "Pod", pod.Namespace+"/"+pod.Name, "Time", time.Now())
+
+	if t.waiting == 0 || pod.Spec.Containers == nil || !containerRequestingTPUs(pod.Spec.Containers...) {
+		return
+	}
+	for _, container := range pod.Spec.Containers {
+		if !containerRequestingTPUs(container) {
+			// Skip to the next container
+			continue
+		}
+		if getEnvironmentVariable("TPU_WORKER_ID", container) == "" {
+			// This TPU pod was not intercepted by the webhook
+			return
+		}
+		// Added a TPU worker Pod to the cache, unblock the next Mutate call
+		t.wg.Done()
+		t.waiting -= 1
+		return
+	}
 }
 
 func main() {
--- a/ray-on-gke/tpu/kuberay-tpu-webhook/main.go
+++ b/ray-on-gke/tpu/kuberay-tpu-webhook/main.go
@@ -17,7 +17,6 @@
 package main
 
 import (
-	"context"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
@@ -31,13 +30,18 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	ray "github.com/ray-project/kuberay/ray-operator/apis/ray/v1"
 	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	listersv1 "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 )
@@ -53,9 +57,11 @@ type slice struct {
 
 // TPUWebhookServer is a KubeRay TPU webhook server instance.
 type TPUWebhookServer struct {
-	// client is used to query the API server for Pods.
-	client     kubernetes.Interface
+	// podLister is used to query Pods from an informer cache.
+	podLister  listersv1.PodLister
 	cacheMutex sync.Mutex
+	wg         sync.WaitGroup
+	waiting    int
 }
 
 // patch is a JSON patch describing mutate operation(s) for an incoming object.
@@ -77,9 +83,9 @@ var (
 	ServerKey      string
 )
 
-func NewTPUWebhookServer(client kubernetes.Interface) *TPUWebhookServer {
+func NewTPUWebhookServer(podLister listersv1.PodLister) *TPUWebhookServer {
 	return &TPUWebhookServer{
-		client: client,
+		podLister: podLister,
 	}
 }
 
@@ -479,10 +485,10 @@ func getNextWorkerID(sliceToWorkerIDs map[slice][]int, podSlice slice, namespace
 }
 
 // getSliceToWorkerIDs returns a mapping representing the current RayCluster state of TPU pods using a PodLister
-func getSliceToWorkerIDs(t *TPUWebhookServer, clusterName string, groupName string, namespace string, numOfHosts int32) (map[slice][]int, error) {
+func (t *TPUWebhookServer) getSliceToWorkerIDs(clusterName string, groupName string, namespace string, numOfHosts int32) (map[slice][]int, error) {
 	sliceToWorkerIDs := make(map[slice][]int)
 
-	podsInGroup, err := t.client.CoreV1().Pods(namespace).List(context.TODO(), metav1.ListOptions{LabelSelector: fmt.Sprintf("ray.io/group=%s", groupName)})
+	podsInGroup, err := t.podLister.Pods(namespace).List(labels.SelectorFromSet(labels.Set{"ray.io/group": groupName}))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


ryanaoleary:
I removed a troubleshooting section with the following:
```
## `TPU_WORKER_ID` assigned to multiple TPU workers in slice

### Symptoms
The webhook outputs the error message `Identical TPU_WORKER_ID assigned to multiple TPU workers in slice`.

### Solution #1
The Ray TPU webhook relies on a PodInformer cache to retrieve the current state of TPU worker Pods in a RayCluster and assign `TPU_WORKER_ID`s. This informer is synced prior to each Pod mutation. However, when quickly deleting and and re-creating RayClusters (especially for larger worker groups), it's possible for the PodLister to retrieve stale information and incorrectly assign `TPU_WORKER_ID`s. This issue is more likely to occur with a large number of worker nodes. The easiest solution in this case is to just delete the Ray custom resource and create it again with `kubectl apply`.
```
I can add this section back in as well as the PodInformer logic if we want to keep using the cache rather than querying the API server directly to ensure consistency.

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
```diff
@@ -590,20 +626,28 @@ func (t *TPUWebhookServer) mutatePod(admissionReview *admissionv1.AdmissionRevie
 	chipsPerHost := getNumTPUChipsRequested(containers...)
 	numOfHosts, _ := getNumTPUHostsFromTopology(clusterName, groupName, namespace, topology, chipsPerHost) // ignore error here because topology may not be set yet
 
-	podsInGroup, err := t.podLister.Pods(namespace).List(labels.SelectorFromSet(labels.Set{"ray.io/group": groupName}))
-	if err != nil {
-		return nil, err
+	// Wait for PodInformer cache to update from previous requests or timeout
+	waitTimeout(&t.wg, time.Second*1)
```
andrewsykim, line 630: Is it expected to run waitTimeout twice, here and the if condition right below?

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
```diff
@@ -708,6 +752,19 @@ func init() {
 	klog.InitFlags(nil)
 }
 
+// addPod allows next goroutine to start once the webhook PodInformer cache updates
+func (t *TPUWebhookServer) addPod(obj interface{}) {
+	// It's not guaranteed the webhook replica that admitted the Pod for this event is the same as the current caller (i.e. wg could be 0).
```
andrewsykim, line 757: Since we're not actually waiting for the pod event for the mutated pod, this is effectively adding some arbitrary delay to pod admission, which may be fixing the race condition with TPU_WORKER_ID.

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
ryanaoleary: Oh only the one in the if condition was intended, I thought I removed the above line in the commit. [ebe0f33](https://github.com/GoogleCloudPlatform/ai-on-gke/pull/859/commits/ebe0f3304d2c5ce4fd4d662c7a52ed049000ab32)

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
ryanaoleary: `AddFunc` is only triggered when a Pod with labels `"ray.io/node-type=worker,app.kubernetes.io/created-by=kuberay-operator"` is added. I can change it to check that it's a TPU Pod with the injected env vars before calling `wg.Done()`, but I wanted to err towards releasing the Wait versus continuously blocking. From my manual testing `Timed out waiting for PodInformer AddFunc` was never showing up in the logs.

ryanaoleary: Added logic to check that the Pod in `addPod` is a TPU worker Pod before unblocking the next `Mutate` call:
04bf73c

ryanaoleary: @andrewsykim I went ahead in [047ff5f](https://github.com/GoogleCloudPlatform/ai-on-gke/pull/859/commits/047ff5f39ed78703f2ddfedd92e5f3605212700e) and changed it to check that the Pod admitted to the cache is the last TPU worker Pod mutated by that webhook replica before unblocking the next goroutine. We check this by adding a `lastAdmitted` var to each `TPUWebhookServer` and setting it to `<replicaIndex>-<TPU_WORKER_ID>` which are vars set for both single-host and multi-host TPUs. If the webhook Pod restarts in between Pod admission requests, the value of `lastAdmitted` will be empty and the `addPod` function will be a no-op (i.e. it won't wait for anything). The PodInformer will be initialized again and obtain an up-to-date list of Pods from the API server, so the next `Mutate` call should proceed correctly. Otherwise, each webhook `Mutate` request will wait for the PodInformer cache to update from the previous request before proceeding which should ensure unique `TPU_WORKER_ID`s even when large slice sizes result in PodInformer updates that are slower than the latency between mutating admission requests.

ryanaoleary: From offline discussion, with [f87aec0](https://github.com/GoogleCloudPlatform/ai-on-gke/pull/859/commits/f87aec0ae3476a886d3e4d17e7beb1c408f1ef86) we now check for `<namespace>-<RayCluster name>-<replicaIndex>-<TPU_WORKER_ID>` which should catch all the cases.

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
```diff
@@ -450,35 +459,48 @@ func getReplicaIndex(sliceToWorkerIDs map[slice][]int, clusterName string, group
...
+	klog.V(1).InfoS("getNextWorkerID", "RayCluster", namespace+"/"+podSlice.clusterName, "Worker Group", podSlice.groupName, "replicaIndex", replicaIndex, "TPU_WORKER_ID", tpuWorkerID)
+	return tpuWorkerID, nil
 }
 
 // getSliceToWorkerIDs returns a mapping representing the current RayCluster state of TPU pods using a PodLister
-func getSliceToWorkerIDs(podsInGroup []*corev1.Pod, clusterName string, groupName string, namespace string, numOfHosts int32) (map[slice][]int, error) {
+func (t *TPUWebhookServer) getSliceToWorkerIDs(clusterName string, groupName string, namespace string, numOfHosts int32) (map[slice][]int, error) {
 	sliceToWorkerIDs := make(map[slice][]int)
 
+	podsInGroup, err := t.podLister.Pods(namespace).List(labels.SelectorFromSet(labels.Set{"ray.io/group": groupName}))
```
andrewsykim, line 492: should this also have ray.io/cluster=<cluster-name> in the selector? Otherwise clusters with similar group names would end up in `podsInGroup`

andrewsykim: Ah, I see later in the code we filter out pods not in the same cluster

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
```diff
@@ -590,19 +627,29 @@ func (t *TPUWebhookServer) mutatePod(admissionReview *admissionv1.AdmissionRevie
...
 	}
 	replicaIndex := getReplicaIndex(sliceToWorkerIDs, clusterName, groupName, namespace)
 	podSlice := slice{clusterName, groupName, namespace, replicaIndex, numOfHosts}
-	tpuWorkerID := getNextWorkerID(sliceToWorkerIDs, podSlice, namespace, replicaIndex) // defaults to 0 for single-host
+	tpuWorkerID, err := getNextWorkerID(sliceToWorkerIDs, podSlice, namespace, replicaIndex) // defaults to 0 for single-host
+	if err != nil {
+		return nil, err
+	}
+	// set the unique identifier for the last admitted Pod by this TPUWebhookServer
+	t.lastAdmitted = fmt.Sprintf("%s-%s-%d-%d", namespace, clusterName, replicaIndex, tpuWorkerID)
```
andrewsykim, line 652: Should last admitted be tracked per RayCluster?

ryanaoleary: Since the Pods are admitted in series by each webhook replica I think it works the same to just check that the last admitted Pod has been added to the cache, regardless of RayCluster. If we did it per-RayCluster we'd also have to handle RayCluster deletion to clean up the list of `lastAdmitted` Pods.

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
ryanaoleary: I went ahead and added `ray.io/cluster` to the selector in [d1a44cc](https://github.com/GoogleCloudPlatform/ai-on-gke/pull/859/commits/d1a44cc66cf31b3abf91c72498369d3c098ce572) so that we get a shorter list of Pods, but yeah we were checking for group, cluster, and namespace when iterating through the Pods.

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
```diff
@@ -708,6 +754,59 @@ func init() {
 	klog.InitFlags(nil)
 }
 
+// addPod allows next goroutine to start once the webhook PodInformer cache updates
+func (t *TPUWebhookServer) addPod(obj interface{}) {
```
spencer-p, line 758: This function has a lot of logic and some delicate conditions to trigger calling wg.Done(), is there a way we can simplify the logic with helper functions? For both readability & risk of breaking the wg calls later

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
```diff
@@ -708,6 +754,59 @@ func init() {
...
+	pod := obj.(*corev1.Pod)
+	klog.V(1).InfoS("addPod", "Pod", pod.Namespace+"/"+pod.Name, "Time", time.Now())
+
+	if t.lastAdmitted == "" {
+		// There is not a pending TPU worker Pod to the informer cache, unblock if waiting and return
+		for {
+			if t.waiting == 0 {
```
spencer-p, line 765: this could be simplified to `for t.waiting > 0`

ryanaoleary: Applied this suggestion in [8f4a349](https://github.com/GoogleCloudPlatform/ai-on-gke/pull/859/commits/8f4a349cc4bc444010085ddcb5bef359c4f154dc).

ray-on-gke/tpu/kuberay-tpu-webhook/main.go
ryanaoleary: Moved all the logic to check if the Pod matches `t.lastAdmitted` to a helper function `isLastAdmittedPod`. I also added a unit test to cover all the cases for addPod. [8f4a349](https://github.com/GoogleCloudPlatform/ai-on-gke/pull/859/commits/8f4a349cc4bc444010085ddcb5bef359c4f154dc)

andrewsykim (approved):

PR #865, opened 2024-10-29
Add active requests Prometheus Metric

Bslabe123:
The following was run to test:

```
python3 benchmark_serving.py --scrape-server-metrics --save-json-results --host=llama3-8b-vllm-service --model=meta-llama/Meta-Llama-3-8B --dataset=ShareGPT_V3_unfiltered_cleaned_split.json --tokenizer=meta-llama/Meta-Llama-3-8B --request-rate="10" --backend=vllm --num-prompts=100 --max-input-length=256 --max-output-length=256 --port=8000
```

Results observed in metrics explorer:
Screenshot 2024-10-29 at 15 56 34
cs explorer:

Note GMP doesn't collect metrics frequently enough to capture decreasing metrics, this was validated manually

### DIFF
--- a/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
+++ b/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
@@ -13,7 +13,7 @@ import random
 import requests
 import time
 from typing import AsyncGenerator, List, Optional, Tuple, Dict
-from prometheus_client import start_http_server, Histogram
+from prometheus_client import start_http_server, Histogram, Gauge
 
 import google.auth
 import google.auth.transport.requests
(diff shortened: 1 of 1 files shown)


benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -211,8 +212,11 @@ async def send_request(
   async with aiohttp.ClientSession(timeout=timeout,trust_env=True) as session:
     while True:
       try:
+        active_requests_metric.inc()
```
achandrasekar, line 215: Shouldn't we count this after the request is posted? Will it be accurate if we do it just before sending the request?

jjk-g: Does aiohttp client session have some maintained prometheus exporter package that makes sense to use in our scenario?

Bslabe123: From what I gather it doesn't, looks like theres a few packages that do this
- https://pypi.org/project/prometheus-async/
- https://pypi.org/project/aiohttp-prometheus-client/

jjk-g: ack, that is fine.

One more thought, is client session able to have custom middleware that might simplify this?

We probably will also need this supported with streaming as well.

Bslabe123: Added

achandrasekar (approved):

PR #870, opened 2024-11-01
Use the tpu_network_optimizer image instead of a raw bash script.

darinpeetz:
The container image contains the script, but makes future updates (if necessary) more straightforward by just pulling a new image rather than updating the sript inside the daemonset. The contents of the image also allow us to mount only /sys and /proc, instead of mounting / entirely.


spencer-p:
Where is this image built from?

Tagging @NoahJohnson1 and @thebinaryone1 to review as they have previously worked on this.

darinpeetz:
Not sure if I'm misunderstanding your question, but the image is built internally as a distroless kubernetes container image.

thebinaryone1 (approved):

PR #871, opened 2024-11-01
GCS Bucket upload support

Bslabe123:
Added the following two flags to benchmark_serving.py:
- `--output-bucket`: upload results in json format to this bucket
- `--output-bucket-filepath`: if specified, upload results to this path within bucket, otherwise upload to root

Requisite changes to terraform also included

### DIFF
--- a/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
+++ b/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
@@ -796,7 +796,8 @@ if __name__ == "__main__":
       "Specifies the destination path within the bucket provided by"
       " --output-bucket for uploading the JSON results. This argument requires"
       " --output-bucket to be set. If not specified, results will be uploaded "
-      " to the root of the bucket."
+      " to the root of the bucket. If the filepath doesnt exist, it will be"
+      " created for you."
     )
   )
   parser.add_argument(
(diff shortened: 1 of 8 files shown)


benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -610,6 +617,19 @@ async def main(args: argparse.Namespace):
...
+  # Should fail fast if client is misconfigured or missing permissions
+  if args.output_bucket is not None:
+    global gcs_client
+    gcs_client = storage.Client()
+    global gcs_bucket
+    gcs_bucket = gcs_client.bucket(args.output_bucket)
+
+    if args.output_bucket_filepath:
+      blob = gcs_bucket.blob(args.output_bucket_filepath)
+      if not blob.exists():
```
annapendleton, line 631: Just double checking my understanding, this behavior creates the filepath if it doesn't exist yet?

annapendleton: If so, we might want to document that as well in the variable definition - eg. "if file path doesn't exist yet, it will create it for you'

Bslabe123: Correct, added!

PR #880, opened 2024-11-14
Make Prompts Reusable for Very Large Experements

Bslabe123:
We currently cannot run experements where the number of prompts exceeds the size of the dataset (for us ~30k prompts), this change makes it so we now instead just filter the dataset and then select randomly from it.

### DIFF
--- a/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
+++ b/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
@@ -52,9 +52,8 @@ trace_config.on_request_end.append(on_request_end)
 gcs_client = None
 gcs_bucket = None
 
-def sample_requests(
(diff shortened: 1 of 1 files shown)


benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -428,6 +425,7 @@ async def benchmark(
         )
       )
     tasks.append(task)
+    prompts_sent += 1
```
annapendleton, line 428: Is the async for loop thread safe? Do we need a lock on this prompts_sent?

annapendleton: Read into the library a bit more, it looks like this should work as is - eg async for loop is iterative and will wait for the next generated request before moving on

PR #899, opened 2024-12-03
Fix TPOT calculation for streaming requests in benchmarking scripts

Edwinhr716:
TPOT should exclude TTFT from its calculation, as the time to first token is higher than that of subsequent tokens.


Edwinhr716:
TPOT will be renamed in #900, closing this PR

PR #908, opened 2024-12-10
Request rate function

Bslabe123:
Request rate flag extended to accept a string containing some `f(t)` where t is time since the start of the benchmarking run. Example inputs: 
- Linear growth: `24+(t/2)`
- Stepped growth (step from 50 to 150 at t=300, do not step up more): `50+100(min(1,(floor(t/300))))`


igooch:
@Bslabe123 is the PR still needed? If not please close PR.

If yes, could you please add someone for your team with context as reviewer?

PR #932, opened 2025-01-15
Make benchmarking script request timeout configurable via `--request-timeout`

Bslabe123:


benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py
```diff
@@ -27,7 +27,7 @@
 from google.protobuf.timestamp_pb2 import Timestamp
 
 MIN_SEQ_LEN = 4
-CLIENT_TIMEOUT_SEC = 3 * 60 * 60
+CLIENT_TIMEOUT_SEC = 10 * 60
```
annapendleton, line 30: let's make this a configurable parameter rather than hardcoding this :)

annapendleton, line 30: We should remove this line as we're using the variable

annapendleton (approved):

PR #939, opened 2025-01-17
Clarify container-image should have a tag in disk image building

raushan2016:
`ctr image pull` only works with image with tag and will fail for images with tags.


vivianrwu:
lgtm!

vivianrwu (approved):

PR #967, opened 2025-02-06
WIP: TPU Provisioner: Node pool hash comparison

nstogner:
Issue:

Today the provisioner just checks if a node pool with the expected name exists for a given workload. If the workload is quickly recreated with a different node selector it is possible that the provisioner will not delete the old node pool in time and it will not think it needs to create a new one since the old one exists.

Fix:

This PR introduces new logic into the provisioner to check that the existing node pool matches the desired node pool via a hash comparison. Hash is calculated at node pool creation time and stored as a node pool label for later comparison.


tpu-provisioner/internal/cloud/gke.go
```diff
@@ -430,7 +485,14 @@ func (g *GKE) nodePoolForPod(name string, p *corev1.Pod) (*containerv1beta1.Node
 		},
 		MaxPodsConstraint: &containerv1beta1.MaxPodsConstraint{MaxPodsPerNode: maxPodsPerNode},
 		NetworkConfig:     networkConfig,
-	}, nil
+	}
+
+	hash, err := nodePoolHash(np)
```
SidneyShen, line 490: This has assumptions that all the constructed node pool is derived from 1. jobset/pod spec and 2. hard-coded section node management policy, upgrade settings. I'm not worried about 1, but for 2, if someone added any changes in this function (accidentally, or default behavior), it will cause a hash diff. and will impact all existing running pools. One option to solve this is probably append a version as part of hash.

nstogner: 1. This should be a valid assumption based on the call stack which is filtered here: https://github.com/GoogleCloudPlatform/ai-on-gke/blob/c44682d1b252afc35ffbdccebbc016b7389c601c/tpu-provisioner/internal/controller/creation_controller.go#L104
2. Great callout, I was originally worried about overzealous node pool recreation under similar circumstances, however, I think this is not a concern since we only call into this codepath [when a Pod is Unschedulable](https://github.com/GoogleCloudPlatform/ai-on-gke/blob/c44682d1b252afc35ffbdccebbc016b7389c601c/tpu-provisioner/internal/controller/creation_controller.go#L107). Perhaps there are some edge cases where a Pod is marked as Unschedulable for a temporary reason unrelated to node pool traits.

SidneyShen: > I think this is not a concern since we only call into this codepath [when a Pod is Unschedulable](https://github.com/GoogleCloudPlatform/ai-on-gke/blob/c44682d1b252afc35ffbdccebbc016b7389c601c/tpu-provisioner/internal/controller/creation_controller.go#L107)

Isn't the other two conditions also trigger the pod reconciliation? https://github.com/GoogleCloudPlatform/ai-on-gke/blob/c44682d1b252afc35ffbdccebbc016b7389c601c/tpu-provisioner/internal/controller/creation_controller.go#L105-L106 

I am thinking the scenario, jobset `foo` in running on node pool `foo-np`, tpu-provisioner is restarted with a new version with new node pool traits, `foo` restarts -> all old jobs deleted and new jobs created, will this cause a `foo-np` recreation with the new hash?

nstogner: Resolved convo offline - Pod must be in unschedulable state, however that state could be reached during a node pool repair so it makes sense to be more careful about node pool recreation.

tpu-provisioner/internal/cloud/gke.go
```diff
@@ -555,3 +617,29 @@ func getAnnotation(p *corev1.Pod, key string) string {
...
+// where values might change due to a config or code change in the provisioner.
+// Example scenario where selective hashing is useful:
+// 1. Provisioner is updated to include new upgrade settings.
+// 2. Some node pool goes into a repairing state.
+// 3. The workload Pod goes into an unschedulable state.
+// 4. The code path for ensuring a matching node pool exists is executed.
+func nodePoolSelectiveHash(np *containerv1beta1.NodePool) (string, error) {
+	h := fnv.New32a()
+	npToHash := &containerv1beta1.NodePool{
+		Config: &containerv1beta1.NodeConfig{
```
SidneyShen, line 632: Didn't include ReservationAffinity: reservation, that means there will be no cases of flipping between reservation and on-demand (at least in the near term), which are treated as same tier today.

SidneyShen (approved):

tpu-provisioner/internal/cloud/gke.go
nstogner: Good catch

nstogner:
Updated to include an interface for interacting with GKE NodePool API in order to add some tests.

tpu-provisioner/internal/cloud/gke.go
```diff
@@ -199,45 +213,80 @@ func (g *GKE) DeleteNodePool(name string, eventObj client.Object, why string) er
...
 
-func (g *GKE) nodePoolForPod(name string, p *corev1.Pod) (*containerv1beta1.NodePool, error) {
+func nodePoolsMatch(desired, existing *containerv1beta1.NodePool) (bool, error) {
+	desiredHash, err := nodePoolSelectiveHash(desired)
+	if err != nil {
+		return false, fmt.Errorf("hashing node pool: %w", err)
+	}
+	if existing.Config != nil && existing.Config.Labels != nil {
+		existingHash, ok := existing.Config.Labels[LabelNodePoolHash]
+		if !ok {
```
echiugoog, line 279: Would it make sense to make this behavior configurable? I imagine over time this case would happen less and less? 

I'd probably want to log a message as well to help with troubleshooting

tpu-provisioner/internal/cloud/gke.go
echiugoog: Would spot be useful as well to include?

echiugoog (approved):
LGTM

tpu-provisioner/internal/cloud/gke.go
nstogner: `Spot:        np.Config.Spot,` ... it is

roberthbailey:
@nstogner - Is this PR still a WIP or should the title be updated?

nstogner:
> @nstogner - Is this PR still a WIP or should the title be updated?

Updated title

SidneyShen (approved):

tpu-provisioner/internal/cloud/gke.go
```diff
@@ -560,3 +639,40 @@ func getAnnotation(p *corev1.Pod, key string) string {
...
+
+// nodePoolSelectiveHash attempts to hash information specific to workload requirements.
+// A selective approach is taken to avoid overzealous node pool recreation under circumstances
+// where values might change due to a config or code change in the provisioner.
+// Example scenario where selective hashing is useful:
+// 1. Provisioner is updated to include new upgrade settings.
+// 2. Some node pool goes into a repairing state.
+// 3. The workload Pod goes into an unschedulable state.
+// 4. The code path for ensuring a matching node pool exists is executed.
+func nodePoolSelectiveHash(np *containerv1beta1.NodePool) (string, error) {
```
Ethanlm, line 651: Should it just return the hash value if it already exists? Then `nodePoolsMatch` can call this function for desired and existing node pool to get the hash

tpu-provisioner/internal/cloud/gke.go
```diff
@@ -560,3 +639,40 @@ func getAnnotation(p *corev1.Pod, key string) string {
...
+				continue
+			}
+			labels[k] = v
+		}
+	}
+	npToHash := &containerv1beta1.NodePool{
+		Config: &containerv1beta1.NodeConfig{
+			Spot:        np.Config.Spot,
+			Labels:      labels,
+			Taints:      np.Config.Taints,
```
Ethanlm, line 668: If a node pool is cordoned, we don't necessary want to create a new node pool for the job, as it will violate Bastion logic. 

The problem we had is pods couldn't be scheduled because it has a node-selector that is different from the labels of existing node pools.  `Spot` information is also in the labels. I think we should only use `labels`.

Ethanlm (commented):
Could you pinpoint where the second issue is fixed in the code?
I am not sure if I understand the second issue :(

tpu-provisioner/internal/controller/deletion_controller.go
```diff
@@ -147,6 +149,7 @@ func (r *DeletionReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&corev1.Node{}).
+		Watches(&corev1.Pod{}, handler.EnqueueRequestsFromMapFunc(handler.MapFunc(nodeForPod))).
```
nstogner, line 152: @Ethanlm Here is where we add the Pod watch back in:

`Pod terminates (we watch pod deletion event) --> We map Pod to its Node --> We reconcile on that Node --> We find the node pool for that node and delete if needed`

nstogner:
> Could you pinpoint where the second issue is fixed in the code? I am not sure if I understand the second issue :(

https://github.com/GoogleCloudPlatform/ai-on-gke/pull/967/files#r1973622851

tpu-provisioner/internal/cloud/gke.go
Ethanlm: I see.

Ethanlm (approved):

Issue #884, opened 2024-11-18
[benchmarking] profile generator metrics scraping improved resilience


annapendleton:
Currently filters on scraping workload metrics relies on the following conditions:

- single pod-monitoring object with "$BACKEND-podmonitoring" across all clusters in the project - https://github.com/GoogleCloudPlatform/ai-on-gke/blob/ee94bdccf6b5e0eef9c8df6c7a44a027250986d1/benchmarks/benchmark/tools/profile-generator/container/benchmark_serving.py#L588

- single workload on the cluster tied to the above pod-monitoring object

We should make this filtering more specific to the desired workload in case there are additional pod monitoring objects in the same project, + if there are multiple pods that the pod monitoring object will relate to on the given cluster

PR #976, opened 2025-02-11
Correctly filter metrics based on namespace and pod label

liu-cong:
Fixes https://github.com/GoogleCloudPlatform/ai-on-gke/issues/884


liu-cong:
This was tested in my cluster. However I have not made the terraform changes.

PR #981, opened 2025-02-16
feat: NV NIM Drug discovery blue print support for L4

sujituk:
- Added support for L4 GPU
 - Caching instruction for AlphaFold2


sujituk:
@roberthbailey if you can take this up. It's similar to last week's [NIM PR](https://github.com/GoogleCloudPlatform/ai-on-gke/pull/948) with updates to ReadMe and file rename

roberthbailey (approved):

PR #989, opened 2025-02-19
Add more metrics

liu-cong:
- **Correctly filter metrics based on namespace and pod label**
- **Add counter type metrics, and add more vllm metrics**


achandrasekar:
Looks like this is marked as a draft. Let me know once it is ready to review.

PR #991, opened 2025-02-21
Qss poc

danjuan-81:
The changes:
1) UI changes based on description doc
2) Moved the recipe templates under the machine type.
3) Fixed template error


applications/hcc/variables.tf
```diff
@@ -60,21 +60,25 @@ variable "a3_ultra_zone" {
 variable "node_count_gke" {
   description = "Toolkit deployment variable: node_count_gke"
   type        = number
+  default     = 0
```
ACW101, line 63: Do we need these default?

ACW101 (approved):

applications/hcc/variables.tf
danjuan-81: Let me remove the default value. It will be safer if there are no default value.

PR #1004, opened 2025-02-28
Add A3Ultra Llama3.1-7b and Llama3.1-70b recipe

danjuan-81:
1) Add A3Ultra Llama3.1-7b and Llama3.1-70b recipe
2) Make reservation name required
3) Add Llama3.1-7b and Llama3.1-70b deployment option back

E2E test from MK UI: https://pantheon.corp.google.com/products/solutions/deployments/details/us-central1/gke-llama-testing?e=13803378&inv=1&invt=Abqvcw&mods=monitoring_api_prod&project=supercomputer-testing


applications/hcc/modules/nemo/main.tf
```diff
@@ -27,24 +27,39 @@ resource "helm_release" "nemo" {
...
     value = var.checkpoint_bucket
   }
 
+  set {
+    name = "volumes.gcsMounts[0].bucketName"
+    value = var.checkpoint_bucket
+  }
+  
+  set {
+    name = "clusterName"
```
ACW101, line 54: Looks like we don't need cluster name anymore since you deleted it

applications/hcc/modules/nemo/recipe.tf
```diff
@@ -4,4 +4,9 @@ locals {
     "llama3.1_70b_nemo_pretraining"  = "llama3-70b-fp8.yaml"
     "gke-nccl"  = ""
   }[var.recipe]
+
+  workload_image = {
+    "A3 Mega"  = "us-central1-docker.pkg.dev/deeplearning-images/reproducibility/pytorch-gpu-nemo@sha256:7a84264e71f82f225be639dd20fcf9104c80936c0f4f38f94b88dfb60303c70e"
```
ACW101, line 9: I wonder if the same image will work for bth a3mega and a3ultra, can you give it a try?

ACW101 (approved):

applications/hcc/modules/nemo/recipe.tf
danjuan-81: Will try.

applications/hcc/modules/nemo/main.tf
danjuan-81: I see. I updated the network name to vpc1..9 to be consistent. Will remove it.

PR #1018, opened 2025-03-07
Adds sample-public-image- prefix to jupyterhub-authentication-class/jupyter-auth-class image tag

igooch:
Add prefix `sample-public-image-` to the image tag to indicate image is not production ready.


gongmax (approved):

gongmax:
There are violations for the shipshape check

PR #1019, opened 2025-03-07
NV NIM Blueprint: Digital human for customer services

sujituk:
Added NIM blueprint support for Digital Human service
   - Audio2face-3D
   - Audio2face-2D
   - FastPitch-hifigan-tts
   - Llama3-8b-instruct
   - nv-embedqa-e5-v5
   - nv-rerankqa-mistral4b-v3
   - Parakeet-ctc-1.1b-asr


tutorials-and-examples/nvidia-nim/blueprints/README.md
```diff
@@ -8,3 +8,14 @@ Here you will find the NVIDIA NIM Blueprints that can be provisioned to run on G
    - DiffDock
 
    You can follow the detailed steps [here](./drugdiscovery/README.md).
+
+2. [Digital Human for Customer Service](https://build.nvidia.com/nvidia/digital-humans-for-customer-service)
```
roberthbailey, line 12: I'd suggest putting this one first, given that I think it will have more general applicability than the bionemo one.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+    export CLUSTER_NAME="gke-nimbp-dighuman"
+    export NP_NAME="gke-nimbp-dighuman-gpunp"
+
+    export ZONE="us-west4-a"            # e.g., us-west4-a
+    export NP_CPU_MACHTYPE="e2-standard-2" # e.g., e2-standard-2
+    export NP_GPU_MACHTYPE="g2-standard-96" # e.g., a2-ultragpu-1g
+
+    export ACCELERATOR_TYPE="nvidia-l4"     # e.g., nvidia-a100-80gb
+    export ACCELERATOR_COUNT="8"            # Or higher, as needed
```
roberthbailey, line 39: can you go higher than 8?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
```diff
@@ -0,0 +1,140 @@
...
+    export NODE_POOL_NODES=1                # Or higher, as needed
+
+    export NGC_API_KEY="<NGC API Key>"
+
+    ```
+
+2. **GKE Cluster creation**:
```
roberthbailey, line 46: and node pool creation

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+    ```bash
+
+    gcloud container clusters create "${CLUSTER_NAME}" \
+      --num-nodes="1" \
+      --location="${ZONE}" \
+      --machine-type="${NP_CPU_MACHTYPE}" \
+      --gateway-api=standard \
+      --addons=GcpFilestoreCsiDriver,HttpLoadBalancing
```
roberthbailey, line 55: Is the `HttpLoadBalancing` parameter here necessary?

From https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#overview:

"To use Ingress, you must have the HTTP load balancing add-on enabled. GKE clusters have HTTP load balancing enabled by default; you must not disable it."

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+2. **GKE Cluster creation**:
+
+    ```bash
+
+    gcloud container clusters create "${CLUSTER_NAME}" \
+      --num-nodes="1" \
+      --location="${ZONE}" \
+      --machine-type="${NP_CPU_MACHTYPE}" \
+      --gateway-api=standard \
```
roberthbailey, line 54: I can't tell from the docs, but it doesn't look like this is enabled by default - can you confirm?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
```diff
@@ -0,0 +1,140 @@
...
+      --num-nodes="${NODE_POOL_NODES}" \
+      --machine-type="${NP_GPU_MACHTYPE}" \
+      --accelerator="type=${ACCELERATOR_TYPE},count=${ACCELERATOR_COUNT},gpu-driver-version=LATEST" \
+      --placement-type="COMPACT" \
+      --disk-type="pd-ssd" \
+      --disk-size="300GB"
+
+    ```
+
+3. **Deploy NIMs:**
```
roberthbailey, line 70: In the other guide you have Get Cluster Credentials and Set kubectl Alias (Optional) as separate steps but here they are combined. I think it would make sense to keep them separate for consistency.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
```diff
@@ -0,0 +1,140 @@
...
+    |`dighum-audio2face-3d-dd-dd` | 1/1 | Running | 0 |
+    |`dighum-fastpitch-tts-ee-ee` | 1/1 | Running | 0 |
+    |`dighum-maxine-audio2face-2d-ff-ff` | 1/1 | Running | 0 |
+    |`dighum-parakeet-asr-1-1b-gg-gg` | 1/1 | Running | 0 |
+
+4. **Access NIM endpoints**
+
+    ```bash
+
+    SERVICES=$(k get svc | awk '{print $1}' | grep -v NAME | grep '^dighum')
```
roberthbailey, line 107: It's unfortunate that there isn't an easier way to do this :/

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
```diff
@@ -0,0 +1,140 @@
...
+      curl http://${EXTERNAL_IP}/v1/health/ready
+      echo " "
+      echo "----------------------------------"
+    done
+
+    ```
+
+    [Click here if you need HTTPS endpoints](https.md)
+
+## Tear down
```
roberthbailey, line 127: Before tear down, can we add a step that does _something_ (akin to the Test Deployments step in the bionemo guide)?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/digital-human-nimbp.yaml
```diff
@@ -0,0 +1,518 @@
...
+    matchLabels:
+      app: dighum-embedqa-e5v5
+  template:
+    metadata:
+      labels:
+        app: dighum-embedqa-e5v5
+    spec:
+      containers:
+      - name: dighum-embedqa-e5v5
+        # image: nvcr.io/nim/nvidia/nv-embedqa-e5-v5:1.0.1
```
roberthbailey, line 85: curious about this - is there not a tagged version that works?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/digital-human-nimbp.yaml
```diff
@@ -0,0 +1,518 @@
...
+      - name: secret-nvcr
+      nodeSelector:
+        cloud.google.com/gke-gpu: "true"
+      volumes:
+        - name: dshm
+          emptyDir:
+            medium: Memory
+            sizeLimit: 16Gi
+
+# ---
```
roberthbailey, line 423: Why is this service commented out? If we don't need it, should it be removed?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/digital-human-nimbp.yaml
```diff
@@ -0,0 +1,518 @@
...
+            mountPath: /dev/shm
+        env:
+        - name: NGC_API_KEY
+          valueFrom:
+            secretKeyRef:
+              name: ngc-api-key
+              key: NGC_API_KEY
+        - name: NIM_HTTP_API_PORT
+          value: "8000"
+        # livenessProbe:
```
roberthbailey, line 483: I'm also curious about this - does the liveness probe not work?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
```diff
@@ -0,0 +1,143 @@
...
+## Prerequisites
+
+- **kubectl:**  kubectl command-line tool installed and configured.
+- **GKE credentials** Ensure you have the credentials to access the GKE cluster.
+- **Certificates**  Privileges to create certificates.
+- **IP Reservation** Privileges to reserve IP addresses.
+
+## Setup
+
+1. **Environment setup**: You'll set up several environment variables to make the following steps easier and more flexible. These variables store important information like cluster names, machine types, and API keys. You need to update the variable values to match your needs and context.
```
roberthbailey, line 19: In this guide it isn't "several" it's "a couple" of environment variables.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
```diff
@@ -0,0 +1,143 @@
...
+
+    ```bash
+
+    export NIMS="dighum-embedqa-e5v5 dighum-llama3-8b dighum-rerankqa-mistral4bv3"
```
roberthbailey, line 23: This looks like a subset of the NIMs deployed; why is that? Do we not need external connectivity for all of them?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
sujituk: Confirmed, its disabled if not added as add-on. I will move the gateway-api and addons to https.md as the cluster can be updated using `gcloud container clusters update`

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
sujituk: done. made separate steps for all - get creds, alias and secrets.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
sujituk: +1

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/digital-human-nimbp.yaml
sujituk: typo, reverted to tagged

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/digital-human-nimbp.yaml
sujituk: the pod logs shows healthy, but when pinged using port-fwd or through LB, it doesn't give the required json.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
sujituk: updated

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
sujituk: good catch. added remaining.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/digital-human-nimbp.yaml
sujituk: underlying pod reported healthy, but the endpoint not giving a OK through browser or port-fwd.

sujituk (commented):
comments addressed.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
sujituk: The prior steps checks and list the service using `curl` to check health status. Is it different from that?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
```diff
@@ -0,0 +1,156 @@
...
+- **GKE credentials** Ensure you have the credentials to access the GKE cluster.
+- **Certificates**  Privileges to create certificates.
+- **IP Reservation** Privileges to reserve IP addresses.
+
+## Setup
+
+1. **Cluster update**: You'll need to update the cluster by enabling GKE Gateway controller and addon for HTTP LoadBalancing
+    ```bash
+
+    gcloud container clusters update "${CLUSTER_NAME}" \
```
roberthbailey, line 22: Can you do both of these updates in a single call? Or does it require two consecutive calls?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
roberthbailey: This checks if the pod is reporting healthy (which should already be verified because otherwise the liveness probes would be failing). I'm wondering if there is anything akin to sending a test prompt to an LLM that verifies each NIM is "working as expected". Or is the health check sufficient for that purpose?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/digital-human-nimbp.yaml
roberthbailey: Can we report this back to nvidia? My understanding is that all NIMs are supposed to have a consistent health status interface so if this one is behaving differently that seems like a bug.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
```diff
@@ -0,0 +1,156 @@
...
+      --location="${ZONE}"  \
+      --gateway-api=standard
+
+    gcloud container clusters update "${CLUSTER_NAME}" \
+      --location="${ZONE}"  \
+      --update-addons=HttpLoadBalancing=ENABLED
+
+    ```
+
+2. **Environment setup**: You'll set up a couple of environment variables to make the following steps easier and more flexible. These variables store important information like cluster names, machine types, and API keys. You need to update the variable values to match your needs and context.
```
roberthbailey, line 32: These variables store... is copied from the other page and isn't relevant here. On the other hand, it would be great to give some guidance as to what I should stick in the `DOMAIN` variable. 

Also, the NIMs variable shouldn't be updated to match my needs / context. It's based on what needs to be exported, right? So it's really only the DOMAIN variable that needs to be customized here.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
```diff
@@ -0,0 +1,156 @@
...
+
+    ```bash
+
+    for NIM in ${NIMS}; do
+      gcloud compute addresses create ${NIM}-ip --global;
+    done
+
+    ```
+
+4. **DNS**: Configure the DNS subdomains for each NIM. Our sub-domains for this example should be in this format <NIM>.<DOMAIN> (e.g. llama3-8b.example.com).
```
roberthbailey, line 51: Is this completely out of band? In the past I've used services like nip.io to create "DNS names" that map to a static IP for the purposes of testing and using temporary certs without having to figure out how to configure DNS subdomains.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/digital-human-nimbp.yaml
sujituk: agreed. will do. can we do this a separate issue? given the gtc time, i dont know if we will have immediate solution.

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
sujituk: @ggiovanejr - can you pls respond to this?

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
sujituk: will do

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
sujituk: @ggiovanejr pls look

ggiovanejr: We can remove it, the variables are only DOMAIN (can be removed) and NIMS (it will have fixed values, right?)

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
ggiovanejr: 100% agree, if you are going to use only nip.io, we don't need DOMAIN and let's keep it simple.

ggiovanejr:
We can remove DOMAIN if only nip.io will be used

ggiovanejr (approved):
LGTM

ggiovanejr (changes requested):
I can see the first 3 services ($NIMS variable) with services on port 8000 and it will work fine, however other services are using port 9000 and also have grpc ports, we have to change the services and HTTPS configuration for them.
@roberthbailey @sujituk

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/README.md
sujituk: added steps to test inference

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/digital-human-nimbp.yaml
sujituk: submitted https://github.com/NVIDIA-AI-Blueprints/digital-human/issues/28

tutorials-and-examples/nvidia-nim/blueprints/digitalhuman/https.md
sujituk: needs 2 separate calls.

sujituk:
@roberthbailey comments addressed. ready for review.

roberthbailey (approved):

PR #1043, opened 2025-03-18
add cloud build check for qss-poc branch

Yevet:


cloudbuild.yaml
```diff
@@ -115,6 +123,52 @@ steps:
...
+    # --project $PROJECT_ID
+
+    cd /workspace/applications/hcc/
+    terraform apply \
+    -var-file=a3ultra_workloads.tfvars \
+    -auto-approve -no-color
+
+    terraform apply \
+    -var-file=a3mega_workloads.tfvars \
+    -auto-approve -no-color
```
danjuan-81, line 147: It creates a3ultra cluster first, then a3mega cluster, finally cleanup both of them. Have you tested this flow?

When I use the terraform commands in my cloudtop. I have to do create A3ultra cluster, then destroy it. After the cluster is destroyed, then I can create another new cluster.

Yevet: solved by only test a3ultra cluster

danjuan-81 (approved):

PR #1067, opened 2025-04-10
Configure maintenance exclusion and add cluster-director label 

zhencao-zc:
- Added `maintenance_exclusions` to the `a3-megagpu-cluster` 
- Added the resource label `gke_product_type: cluster-director-qss` to identify the clusters provisioned by Cluster Director QSS

### DIFF
--- a/applications/hcc/modules/embedded/modules/scheduler/gke-cluster/main.tf
+++ b/applications/hcc/modules/embedded/modules/scheduler/gke-cluster/main.tf
@@ -16,7 +16,12 @@
 
 locals {
   # This label allows for billing report tracking based on module.
-  labels = merge(var.labels, { ghpc_module = "gke-cluster", ghpc_role = "scheduler" })
+  labels = merge(var.labels, { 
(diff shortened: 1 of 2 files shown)


applications/hcc/modules/embedded/modules/scheduler/gke-cluster/main.tf
```diff
@@ -16,7 +16,12 @@
 
 locals {
   # This label allows for billing report tracking based on module.
-  labels = merge(var.labels, { ghpc_module = "gke-cluster", ghpc_role = "scheduler" })
+  labels = merge(var.labels, { 
+    ghpc_module = "gke-cluster", 
+    ghpc_role = "scheduler",
+    "gke_product_type" = "cluster-director-qss"
```
ACW101, line 22: nit: don't need quote around "gke_product_type"

zhencao-zc: Done!

PR #1068, opened 2025-04-10
Integrate ray on hcc qss (backend)

Yevet:
- add ray operator and ray cluster option to hcc (backend logic only)


fcabrera23:
@Yevet  pleae move this PR to the new https://github.com/ai-on-gke/quick-start-guides

PR #1078, opened 2025-04-15
Remove jupyter QSS e2e test

gongmax:
**What type of PR is this?**
> Uncomment only one ` /kind <>` line, press enter to put that in a new line, and remove leading whitespace from that line:
>
> /kind breaking
> /kind bug

/kind cleanup

> /kind documentation
> /kind enhancement

**What this PR does / Why we need it**:
We've already removed the Jupyter QSS application code in https://github.com/GoogleCloudPlatform/ai-on-gke/pull/1069/files. This PR removes it from the E2E test accordingly.

**Which issue(s) this PR fixes**:
CI failures like https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/ae8dffdb-fe62-4e11-ba44-949dcd7fb0d2;step=2?e=13802955&hl=en&invt=Abu1Zw&mods=logs_tg_staging&project=gke-ai-eco-dev

**Special notes for your reviewer**:

PR #1086, opened 2025-04-22
Moving RAG to the new AI on GKE repo

fcabrera23:
**What type of PR is this?**
> /kind cleanup

**What this PR does / Why we need it**:
We're moving all the documentation of the RAG on GKE application into the new quick-start-solutons repository under the AI on GKE org. All content will be shared to customers using the new GKE AI Labs repository.


gongmax (commented):
We need similar change as https://github.com/GoogleCloudPlatform/ai-on-gke/pull/1078/files to prevent CI failures

cloudbuild.yaml
```diff
@@ -129,91 +121,6 @@ steps:
...
-    kubectl exec -it -n rag-$SHORT_SHA-$_BUILD_ID jupyter-admin -c notebook -- jupyter nbconvert --to script /data/rag-kaggle-ray-sql-interactive.ipynb
-    kubectl exec -it -n rag-$SHORT_SHA-$_BUILD_ID jupyter-admin -c notebook -- ipython /data/rag-kaggle-ray-sql-interactive.py
-
-    python3 ./applications/rag/tests/test_rag.py "http://127.0.0.1:8081/prompt"
-    echo "pass" > /workspace/rag_prompt_result.txt
-
-  allowFailure: true
-  waitFor: ['create gke cluster']
-
 - id: 'Generate Kubeconfig'
```
igooch, line 124: It looks like this step and the following steps / tests are running on the rag cluster. Do we still need these tests?

igooch: Adding the note that is looks like the same cluster is used for ray and rag, so these may still be needed until ray and the cluster creation are also removed.

igooch:
The test is failing at this line https://github.com/GoogleCloudPlatform/ai-on-gke/blob/df008358568de4bcb2b6a691cceb93023cc0d9fe/cloudbuild.yaml#L97

As a test you could potentially bump the wait time to see if it's simply an issue of the pods becoming ready.

Other options are to see if there's any code that was ripped out that is critical to the ray test, or to simply rip out both the ray and rag tests as part of the same PR.

fcabrera23:
Deleted cloudbuild.yaml, as this will be part of the new repo, and both RAG and Ray are being migrated

gongmax:
I'm merging this now. I will monitor the future PR and disable the pr review trigger for the RAG/Ray e2e test if needed.

